PR #1, opened 2019-07-25
Remove RawWindowHandleEx

kvark:
I see the value in traits that abstract away multiple implementations. In this case, there is a custom trait per platform. Is the original idea that winit/SDL/GLFW would implement those traits? If so, is there a downside for them to just return `RawWindowHandle` to begin with?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,10 @@
 [package]
 name = "raw-window-handle"
 version = "0.1.0"
-authors = ["Osspial <osspial@gmail.com>"]
+authors = [
+	"Osspial <osspial@gmail.com>",
+	"Dzmitry Malyshau <kvarkus@gmail.com>",
+]
 edition = "2018"
 description = "Raw platform window types for interoperability"
 license = "Unlicense"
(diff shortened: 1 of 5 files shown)


Osspial:
`RawWindowHandleExt` isn't a trait I expect downstream crates to implement. Instead, it's following a design precedent set by the standard library that I quite like - when methods are only available on certain platforms, force users to explicitly import those methods by locking them behind a trait (e.g. [`FileExt`](https://doc.rust-lang.org/std/os/windows/fs/index.html)). The `Ext` suffix might be misleading here since it's core functionality instead of extension functionality, but I still like the design pattern in general.

kvark:
>  force users to explicitly import those methods by locking them behind a trait

I agree it's nice for `FileExt`. With `File`, you use the portable stuff in most cases, and opt-into the platform-specific stuff, thus the trait is useful to separate those parts. For `RawWindowHandle` there is no former part ("portable" stuff), so there is no point to have a separation in terms of the trait.

Ralith:
> design precedent set by the standard library

I believe this is considered a deprecated pattern in the standard library, in favor of just `#[cfg(..)]`ing out platform-specific code.

Osspial:
Closing in favor of the design implemented in #6.

Issue #12, opened 2019-08-19
Android Support


msiglreith:
Exposing an enum variant for Android with an `ANativeWindow` would be great!

jchapman127:
This is required to complete moving `gfx-backend-vulkan` over to `raw-window-handle`.

Osspial:
Closing due to #21.

Issue #11, opened 2019-08-15
XCB support


Ralith:
The current implementation treats X11 as a monolith. Per Vulkan convention, the X11 case should probably be renamed Xlib, and an Xcb case added.

msiglreith:
I'm currently in the process of porting ash-window towards 0.2 (XCB and Android support 👍).
While splitting X11 into Xlib and Xcb I noticed that the XcbHandle has a [`surface`](https://docs.rs/raw-window-handle/0.2.0/raw_window_handle/unix/struct.XcbHandle.html#fields), which seems kind of an odd naming as there doesn't seem to be an actual surface type in the xcb api. Vulkan uses the term `window` with type `xcb_window_t` and I believe raw-window-handle assumes takes same thing here.

Osspial:
Hmm. Honestly, I'm not too happy about that, but I should've reviewed the PR more thoroughly. Sorry about that.

Issue #3, opened 2019-07-26
Prefer `core::ffi::c_void` type over `libc::c_void`


Lokathor:
All `c_void` are in some sense similar since you can always just cast between them, however, since core offers a `c_void` type we should at least use that common point of communication as much as possible.

Osspial:
I'm not opposed to doing this, but it's worth noting that we'll still have to pull in `libc` to maintain `no_std` compatibility on Linux, since X11 uses a `c_ulong` to represent the window handle.

PR #17, opened 2019-08-24
Fixes for issue #3, #11 and #12

Didgy74:
Line 36 in src/unix.rs contains the definition for the surface of a Xcb handle. I checked through xproto.h and could only see it being defined as `typedef uint32_t xcb_window_t;` for all platforms. Whether directly defining at as u32 in the Rust code is good enough is up for debate I guess. I personally think it's fine.

### DIFF
--- a/src/android.rs
+++ b/src/android.rs
@@ -1,29 +1,29 @@
 use core::ffi::c_void;
 use core::ptr;
 
-/// Raw window handle for Windows.
+/// Raw window handle for Android.
 ///
 /// ## Construction
 /// ```
-/// # use raw_window_handle::windows::WindowsHandle;
-/// let handle = WindowsHandle {
+/// # use raw_window_handle::android::AndroidHandle;
+/// let handle = AndroidHandle {
 ///     /* fields */
-///     ..WindowsHandle::empty()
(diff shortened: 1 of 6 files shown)


Didgy74:
Update on this pull request. I've now added support for Android support to provide a solution for issue #12. I've also applied the keyword Self where applicable to make for more idiomatic Rust code.

On the Android support, I used the `VkAndroidSurfaceCreateInfoKHR` structure from the Vulkan specification for reference for building the AndroidHandle struct. The definition of this structure can be found a bit lower down on [this page](https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#_wsi_surface).

The Vulkan specification has this to say regarding how to define `ANativeWindow`
```
To remove an unnecessary compile-time dependency, an incomplete type definition of ANativeWindow is provided in the Vulkan headers:

struct ANativeWindow;

The actual ANativeWindow type is defined in Android NDK headers.
```
So I used `*mut c_void` for defining the `ANativeWindow*` field. More elegant solutions are welcome.

Didgy74:
I've split this PR up into multiple ones. Closing the PRs.

PR #20, opened 2019-08-27
Added support for Xcb, in response to issue #11.

Didgy74:
Separated the Xcb changes from PR #17.


Didgy74:
This is an interface breaking change, due to X11 being renamed to Xlib. Consider bumping version.

Osspial:
Since this is a breaking change, I'm holding off on merging this until we finalize discussion on #23.

Issue #16, opened 2019-08-24
Can we remove most of the conditional compilation stuff?


Lokathor:
The conditional compilation stuff is an unnecessary complication to the system.

Basically, every library that _consumes_ a `HasRawWindowHandle` value already needs to have a default, catch-all case where they get some unknown handle type and then error out. We even want the enum to be tagged as non-exhaustive so that the library can be "future proof" and force people to have a default cause so that we don't break semver when new window systems are eventually added.

So since your consuming code's match statement already needs to say
```rust
  unknown => return Err(format!("unknown window handle type: {:?}", unknown)),
```
at the end of it, then having the IOS handle conditionally compiled to not exist on windows doesn't actually add anything. Either way you can't have exhaustive matching.

And since we're just giving all of our handle definitions in terms of libc types and `core::ffi::c_void` and so forth there's not really a problem with pulling in all sorts of platform dependencies if we just have every variant compiled in on all platforms.

So I think that never doing any conditional compilation is probably better in the long term. Makes things simpler to reason about.

hecrj:
Have we considered removing the unexhaustive `RawWindowHandle` enum and `HasRawWindowHandle` trait altogether and adding platform-specific traits instead?

What is the purpose of giving up guarantees by wrapping the handle in an unexhaustive enum when our users have to implement platform-specific code anyway?

Lokathor:
No the point is that an _end user_ of this communication point sees it as just one thing 

```rust
let surface = instance.make_surface_for_window(&window).unwrap();
```

Since some systems potentially have more than one window system available, for this to move all the complexity away from the end user there cannot be window specific methods.

hecrj:
@Lokathor The same way we currently [force our users to pattern match using conditional compilation](https://github.com/gfx-rs/wgpu/blob/master/wgpu-native/src/instance.rs#L99), we can force them to declare a function using conditional compilation too ([they probably have to do this anyway](https://github.com/gfx-rs/wgpu/blob/master/wgpu-native/src/instance.rs#L147)):

```rust
#[cfg(target_os = "ios")]
fn make_surface_for_window<W: HasIOSHandle>(&self, window: &W) -> Result<Surface, Error> {
    // window.ui_window()
    // window.ui_view()
    // ...
}
```

The main difference is that there won't be any need for non-exhaustive variants or fields, and there won't be misleading APIs taking an unsupported window handle and erroring at runtime.

Lokathor:
[...]

PR #23, opened 2019-08-28
Draft prototype using traits

hecrj:
As discussed in #16, an approach using traits instead of non-exhaustive enums/structs. I also added an example showcasing how a `create_surface` implementation for MacOS and Linux would look with this approach.

This is probably a controversial change. I think we should dismiss this idea if there is no clear consensus.

### DIFF
--- a/examples/create_surface.rs
+++ b/examples/create_surface.rs
@@ -1,0 +1,61 @@
+use raw_window_handle::HasRawWindowHandle;
+
+fn main() {}
+
+pub struct Instance {
+    // ...
+}
+
+pub struct Surface;
+
+pub enum Error {
+    NotSupported,
+}
+
+impl Instance {
+    pub fn create_surface<W: HasRawWindowHandle>(&mut self, window: &W) -> Result<Surface, Error> {
+        #[cfg(target_os = "linux")]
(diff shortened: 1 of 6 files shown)


Osspial:
cc @kvark @Ralith

Looking over the changes for this, I've got a couple points of feedback:

1. The biggest downside to this is that it makes the documentation significantly less readable. There are probably ways we can compensate for that, but I don't think we can get rustdoc to render this as intuitively as we have set up now.
2. I can see the current solution for the X11/Wayland backend on Unix being difficult to implement, since it requires auxiliary types for the `HasX11Handle` and `HasWayland` handle traits. That won't be a problem for Winit, but I assume it'll be more of a hassle for SDL wrappers.

One solution that I believe would largely keep the benefits of both the current design and this redesign would be to expose all the window-handle-getters in the top-level trait, then conditionally compile each of those functions so that they're only available on their target platform. I'd write up a sketch of that design, but I just looked back through the original thread and... [it looks like that's pretty much the original design I proposed](https://github.com/rust-gamedev/wg/issues/26#issuecomment-514050190). Welp.

Osspial:
I'm going to close this, since it's not an unambiguously good change to the codebase. Thank you for pulling this together, though!

PR #26, opened 2019-09-03
Web support via stdweb and web-sys

ryanisaacg:
Old PR:
This will probably not be the final change, because it:

- A) will probably require bikeshedding as to the proper way to choose between stdweb / web-sys at compile time
- B) does not implement 'empty'
- C) most importantly, removes 'Copy' from the public API of the RawWindowHandle

Some aspect of web support does need to make it into raw-window-handle for web support in winit, but I wanted to put up this PR to start discussion on what that should look like.

Edit: For more context, the Canvas objects in stdweb and web-sys are Clone but not copy, so a struct containing them may not be Copy either. I'm not sure of the best-practice way to work around this.

---
New PR:

A backend-agnostic numerical ID for web has been added as an option to the window handle enum, with documentation referring users to custom data attributes on HTML elements.

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -12,6 +12,8 @@
 pub struct WebHandle {
     /// An ID value inserted into the data attributes of the canvas element as 'raw-handle'
     ///
+    /// When accessing from JS, the attribute will automatically be called rawHandle
+    ///
     /// Each canvas created by the windowing system shoudl be assigned their own unique ID.
     /// 0 should be reserved for invalid / null IDs.
     pub id: u32,
--- a/src/web.rs
+++ b/src/web.rs
@@ -1,0 +1,31 @@
+/// Raw window handle for the web
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::web::WebHandle;
+/// let handle = WebHandle {
+///     /* fields */
+///     ..WebHandle::empty()
+/// };
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct WebHandle {
+    /// An ID value inserted into the data attributes of the canvas element
+    ///
+    /// Each canvas created by the windowing system shoudl be assigned their own unique ID.
+    /// 0 should be reserved for invalid / null IDs.
+    pub id: u32,
+    #[doc(hidden)]
+    #[deprecated = "This field is used to ensure that this struct is non-exhaustive, so that it may be extended in the future. Do not refer to this field."]
+    pub _non_exhaustive_do_not_use: crate::seal::Seal,
+}
+
+impl WebHandle {
+    pub fn empty() -> WebHandle {
+        #[allow(deprecated)]
+        WebHandle {
+            id: 0,
+            _non_exhaustive_do_not_use: crate::seal::Seal,
+        }
+    }
+}
--- a/src/web.rs
+++ b/src/web.rs
@@ -10,7 +10,7 @@
 /// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct WebHandle {
-    /// An ID value inserted into the data attributes of the canvas element
+    /// An ID value inserted into the data attributes of the canvas element as 'raw-handle'
     ///
     /// Each canvas created by the windowing system shoudl be assigned their own unique ID.
     /// 0 should be reserved for invalid / null IDs.
--- a/src/web.rs
+++ b/src/web.rs
@@ -14,7 +14,7 @@ pub struct WebHandle {
     ///
     /// When accessing from JS, the attribute will automatically be called rawHandle
     ///
-    /// Each canvas created by the windowing system shoudl be assigned their own unique ID.
+    /// Each canvas created by the windowing system should be assigned their own unique ID.
     /// 0 should be reserved for invalid / null IDs.
     pub id: u32,
     #[doc(hidden)]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -46,7 +46,9 @@ pub mod windows;
 #[cfg_attr(feature = "nightly-docs", doc(cfg(target_os = "ios")))]
 #[cfg_attr(not(feature = "nightly-docs"), cfg(target_os = "ios"))]
 pub mod ios;
-// pub mod wasm;
+#[cfg_attr(feature = "nightly-docs", doc(cfg(target_arch = "wasm32")))]
+#[cfg_attr(not(feature = "nightly-docs"), cfg(target_arch = "wasm32"))]
+pub mod web;
 
 mod platform {
     #[cfg(target_os = "macos")]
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -17,6 +17,17 @@ appveyor = { repository = "rust-windowing/raw-window-handle" }
...
+version = "0.3.22"
+optional = true
+features = [
+    'HtmlCanvasElement',
+]
+
+[target.'cfg(target_arch = "wasm32")'.dependencies.stdweb]
+version = "0.4.18"
+optional = true
+
```
Osspial: I'm not entirely comfortable with re-exporting these types in our public API, since it opens us up for breaking changes if the downstream crates ever bump their major version.

Looking at the `stdweb` source code, it looks like the `Canvas` struct just wraps around an `i32` that specifies the `HTMLCanvasElement` JS object. Could we expose that `i32` instead of the higher-level wrappers, than have downstream users reconstruct the object from that integer?

src/web.rs
```diff
@@ -0,0 +1,14 @@
+#[cfg(not(any(feature = "stdweb", feature = "web-sys")))]
+compile_error!("Please choose either the stdweb or web-sys feature when compiling to web");
+
+/// Canvas reference for the web
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct WebHandle {
+    #[cfg(feature = "web-sys")]
+    pub canvas: web_sys::HtmlCanvasElement,
+    #[cfg(feature = "stdweb")]
+    pub canvas: stdweb::web::html_element::CanvasElement,
```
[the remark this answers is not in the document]
ryanisaacg: My reasoning was 'Winit and X11 are both valid choices in the same binary, but stdweb and web-sys are not.' With that in mind, if they should still be separate I have no problem separating them.

Cargo.toml
ryanisaacg: I'll try to look into ways to remove the stdweb and web-sys types from the public API, which should also possibly help with my problem that they aren't 'Copy'.

ryanisaacg: Would using Box::into_raw and Box::from_raw on the canvas objects make sense as a design? Then the handle can just be a `void*`.

The downside is a serious loss of safety, but the upside is that it's more in line with the rest of the API.

Osspial: @ryanisaacg I'm not entirely comfortable with doing that, since it seems like an unnecessary layer of indirection over exposing the JS reference integer directly.

`stdweb`'s `Reference` type exposes unsafe [`from_raw_unchecked`](https://docs.rs/stdweb/0.4.2/src/stdweb/webcore/value.rs.html#27) methods for internal use. Could we see about exposing those methods in the public API? From a light investigation it seems the main reason they aren't exposed is because the authors didn't think there was much reason to expose them (based on https://github.com/koute/stdweb/issues/339), but I think we've got an acceptable reason here.

ryanisaacg: Ok, I'll propose upstream methods to stdweb and wasm-bindgen to have unsafe methods for converting back and forth between the integer ID's.

ryanisaacg: I checked the issue tracker for wasm-bindgen to see if something like this had already been suggested, and it looks like the maintainers are against it: https://github.com/rustwasm/wasm-bindgen/issues/999

Mainly it seems adding the ability to access the integer internals of JsValue would constrain their future designs too strictly.

Osspial: Hmm. It seems we're at an impasse, then - they're unwilling to add `i32` conversions, and we can't add web support without either a) entering a safety minefield if we use `*mut c_void` and they change their internal representation or b) entering a compatibility minefield if either library publishes a breaking change.

I'm not sure how to proceed here. Without upstream support, all the options seem terrible.

ryanisaacg: I think the least-terrible option is to expose the types and re-export web-sys and stdweb, to ensure downstream customers can always use the version being exported by raw-window-handle.

Osspial: Setting ourselves up for either a breaking change or forcing our users to use legacy code would fundamentally undermine the stability and common ground that this library is supposed to provide, and as such is an unacceptable solution.

In the thread you linked, there's a point in @alexcrichton's comment here (https://github.com/rustwasm/wasm-bindgen/issues/999#issuecomment-434818372) which strikes me as strange (emphasis mine):

> I think you're right that taking advantage of future wasm feature "in their fullest" is likely a breaking change, but what I was hoping for is that it's a largely cosmetic breaking change rather than "rewrite the world" breaking change. For example switching some types around is pretty easy, but **losing a fundamental capability (like "O(1) conversion to an integer from JsValue") runs the risk of making it much harder to migrate in the future.**

Wouldn't changing an unsafe `JsValue::from_raw(i32)` method into `JsValue::from_raw(anyref)` just be a cosmetic breaking change? You aren't losing the ability to convert arbitrary integers into `JsValue`, because such an API would never provide that guarantee to begin with. Unsafe code can require additional semantics for proper use, and as long as a `from_raw(i32)` method requires handling the raw `i32` similarly to how you'd handle an `anyref` variable you shouldn't run into any major compatibility issues when you switch the types over.

ryanisaacg: As discussed on the Rust Community discord, if we're willing to limit ourselves to canvases that exist within the DOM, we can provide some sort of UUID that windowing libraries attach to canvas elements. Then, clients can find the canvas via the UUID.

Upsides:
- We're not tied to stdweb *or* web-sys under the hood

Downsides:
- Requires a bit more work on behalf of the windowing libraries (probably not too bad)
- Does not account for canvases not inserted into the DOM (probably not important, because this library is for interop from windowing libraries)

Would this work for everyone?

Osspial: I would be okay with that.

icefoxen: It's kinda a hack IMO, but it's a hack that should work and make it possible to use *right now* without needing to wait for someone else to figure their stuff out.  Which is partially the point of this library.

ryanisaacg:
Ok, I've pushed up a change with a new design that doesn't tie us into the API of web-sys or stdweb.

src/web.rs
```diff
@@ -0,0 +1,31 @@
...
+/// let handle = WebHandle {
+///     /* fields */
+///     ..WebHandle::empty()
+/// };
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct WebHandle {
+    /// An ID value inserted into the data attributes of the canvas element
+    ///
+    /// Each canvas created by the windowing system shoudl be assigned their own unique ID.
```
Osspial: *should

src/web.rs
```diff
@@ -0,0 +1,31 @@
...
+/// ```
+/// # use raw_window_handle::web::WebHandle;
+/// let handle = WebHandle {
+///     /* fields */
+///     ..WebHandle::empty()
+/// };
+/// ```
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct WebHandle {
+    /// An ID value inserted into the data attributes of the canvas element
```
Osspial: This doesn't specify the name of the canvas attribute.

ryanisaacg: Good point, fixed

PR #27, opened 2019-09-26
Release 0.2.0

Osspial:
I'll leave this open for a day or so in case people want to object to the ruling I made in #23, but otherwise this should be GTG.

cc @hecrj @kvark @Didgy74 @Ralith


msiglreith:
Changelog misses Android support \o

Osspial:
@msiglreith Ah, thanks for catching that! I also forgot to bump the version in cargo.toml...

Issue #14, opened 2019-08-21
How should users interface with graphics libraries on macOS/iOS?


seivan:
Say I create the iOS struct, pass in pointers to `UIView` alone, without the rest - because technically they would be unnecessary and unused. 

How would `gfx-hal` know wether or not they need to setup things for Metal vs the `UView` containing a `CAMetalLayer` and etc? 

To clarify, there is no indication mentioned whether or not the related graphics stuff has been setup before hand (e.g using SDL2 to set everything up) vs expecting `gfx-hal` to create the necessary contexts and `CAMetalLayer`?

msiglreith:
Could we rename the topic towards sth like: `MacOS/IOS Interface`?

---

Current interfaces:
```rust
pub struct MacOSHandle {
    pub ns_window: *mut c_void,
    pub ns_view: *mut c_void,
}
pub struct IOSHandle {
    pub ui_window: *mut c_void,
    pub ui_view: *mut c_void,
    pub ui_view_controller: *mut c_void,
}
```
The graphics APIs (Metal and Vulkan Portability) expect a `CAMetalLayer` in their window system interface. Earlier MoltenVK versions exposed an API surface taking NSView/UIView. Are there other relevant graphics APIs which communicate with the views/windows?

At the end of the day I guess it boils down which side (window <-> graphics api) has to do potential conversion steps.

Lokathor:
So when trying to implement this for beryllium yesterday (SDL2 based lib) I was able to get an NS_Window from SDL2 but it doesn't give an NS_View.

Is this field required? Seems like no, because the mac folks told me you can always get a view from the window pointer.

Should we change the Mac handle to be an Option<NonNull<c_void>> with the None version more clearly meaning "get the default yourself" and the Some version being "use this specific sub-view here".

Osspial:
@Lokathor There are some fields on other platforms with similar-ish semantics (e.g. `hinstance` on Windows). I'm more inclined to add the following language to the root `HasRawWindowHandle` docs:

```diff
 /// It is entirely valid behavior for fields within each platform-specific `RawWindowHandle` variant
 /// to be `null` or `0`, and appropriate checking should be done before the handle is used. However,
 /// users can safely assume that non-`null`/`0` fields are valid handles, and it is up to the
 /// implementor of this trait to ensure that condition is upheld.
 ///
+/// Still, implementors should make a best-effort attempt to fill in all available fields. If an
[...]

Issue #2, opened 2019-07-25
Add HINSTANCE for win32


msiglreith:
Vulkan requires for surface creation at least a HWND and HINSTANCE handle.

Osspial:
Is this necessary? I'm pretty sure you can call `GetModuleHandle(ptr::null()) as HINSTANCE` to get the `HINSTANCE` from anywhere in the program.

msiglreith:
IIRC there is some discrepancy when the hwnd comes from an external library or something like that

Osspial:
The only discrepancy I'm aware of is that when you call `GetModuleHandle` from a DLL, the module handle is from the parent executable, not from the DLL. Since Rust links downstream libraries directly into the executable that shouldn't cause any issues. I'd be surprised if that were an issue even with a Rust library compiled to a DLL, but I'd need to test that to know for sure.

Ralith:
> Since Rust links downstream libraries directly into the executable that shouldn't cause any issues

Is this really an assumption we want to bake in? It's perfectly reasonable for people to manually dynamically load libraries containing rust code from both rust and foreign processes. If there was really never any case in which being explicit was necessary, I don't think it would be in the Vulkan interface.

Osspial:
@Ralith Fair enough. I've submitted a PR to add `HINSTANCE` in #30.

PR #28, opened 2019-09-28
Rename XcbHandle surface to window and extend documentation

msiglreith:
Should solve possible confusion due to the usage of base types.

Fixes https://github.com/rust-windowing/raw-window-handle/issues/11

### DIFF
--- a/src/android.rs
+++ b/src/android.rs
@@ -13,6 +13,7 @@ use libc::c_void;
 /// ```
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct AndroidHandle {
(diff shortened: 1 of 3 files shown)


Osspial:
I'll publish a `0.3` shortly. Ideally, this should be our last breaking change.

msiglreith:
https://github.com/rust-windowing/raw-window-handle/issues/2 and https://github.com/rust-windowing/raw-window-handle/issues/14

Are probably points which should be decided on (beforehand?)

Osspial:
#2 and #14 can be resolved without breaking backwards compatibility, so we don't need to block a `0.3` release on them.

PR #33, opened 2019-10-05
Add clause to documentation clarifying the semantics of null values

Osspial:
Addresses https://github.com/rust-windowing/raw-window-handle/issues/14#issuecomment-538686317

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -81,6 +81,11 @@ mod platform {
 /// users can safely assume that non-`null`/`0` fields are valid handles, and it is up to the
 /// implementor of this trait to ensure that condition is upheld.
 ///
+/// Despite that qualification, implementors should still make a best-effort attempt to fill in all
+/// available fields. If an implementation doesn't, and a downstream user needs the field, it should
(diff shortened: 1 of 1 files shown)


Lokathor:
@bvssvni Hello! I'm pinging you about this docs update as the rep for GLFW, but feel free to send whoever else would be better suited as a contact point. Ossipal is reprisenting `winit` interests and I'm speaking for how SDL2 works.

Hopefully this makes your GLFW job easier/clearer!

bvssvni:
@Lokathor I have not followed this discussion, can you fill me in?

Lokathor:
Essentially, we're confirming that if the backend cannot provide a reasonable value it should return a null pointer or 0 (depending on type) instead of a panic. Then any lib that uses this output is implicitly authorized to try and ask the OS for more info if that happens.

Basically, for GLFW, if there's any case that's expensive or impossible to handle, just throw down a null and the consumer can try on their end if it's absolutely necessary.

For example, SDL2 doesn't give an NSView on Mac, so my wrapper lib just gives the NSWindow and a lib calling for the raw window handle can use that to look up the view themselves if it's really necessary.

PR #44, opened 2019-12-18
Add basic support for Solaris

psumbera:
This seems to be the first step to have latest Mozilla Firefox building on Solaris again. Though Solaris doesn't support Wayland.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,7 +31,8 @@ pub mod macos;
         target_os = "dragonfly",
         target_os = "freebsd",
         target_os = "netbsd",
-        target_os = "openbsd"
+        target_os = "openbsd",
+        target_os = "solaris"
     )))
 )]
 #[cfg_attr(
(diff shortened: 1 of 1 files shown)


Osspial:
If Solaris doesn't support Wayland, is there any particular reason to enable it for Solaris?

psumbera:
As far as I know BSD systems also do not support Wayland. I have run into the issue again so I would like to resolve it now:

https://github.com/gfx-rs/gfx/issues/3547

jbeich:
> As far as I know BSD systems also do not support Wayland

DragonFly and FreeBSD support wlroots-based compositors (e.g., sway, hikari, wayfire, phoc, gamescope, river). NetBSD supports swc-based compositors (e.g., velox). OpenBSD has preliminary work on libwayland.

For example, I'm on FreeBSD + Sway where raw-window-handle as used by Alacritty, Veloren, Firefox (`MOZ_WEBRENDER=1`, `MOZ_ENABLE_WAYLAND=1`) works fine.

See also https://lobste.rs/s/awzk3c/wayland_on_netbsd_trials_tribulations

Lokathor:
This is stalled, and i'm just going to merge it, and then we can re-enable wayland later if we need to.

Lokathor (approved):

PR #47, opened 2020-01-07
Initial redox support.

FloVanGH:
Initial support for Redox OS https://redox-os.org/.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -24,6 +24,9 @@ pub mod ios;
 #[cfg_attr(feature = "nightly-docs", doc(cfg(target_os = "macos")))]
 #[cfg_attr(not(feature = "nightly-docs"), cfg(target_os = "macos"))]
 pub mod macos;
+#[cfg_attr(feature = "nightly-docs", doc(cfg(target_os = "redox")))]
+#[cfg_attr(not(feature = "nightly-docs"), cfg(target_os = "redox"))]
+pub mod redox;
 #[cfg_attr(
     feature = "nightly-docs",
     doc(cfg(any(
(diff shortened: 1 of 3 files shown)


src/lib.rs
```diff
@@ -105,6 +110,10 @@ pub enum RawWindowHandle {
     #[cfg_attr(not(feature = "nightly-docs"), cfg(target_os = "macos"))]
     MacOS(macos::MacOSHandle),
 
+    #[cfg_attr(feature = "nightly-docs", doc(cfg(target_os = "redox")))]
+    #[cfg_attr(not(feature = "nightly-docs"), cfg(target_os = "redox"))]
+    MacOS(redox::RedoxHandle),
```
Osspial: This variant should be named `RedoxHandle`.

src/redox.rs
```diff
@@ -0,0 +1,31 @@
+use core::ptr;
+use orbclient::Window;
```
Osspial: Part of `raw-window-handle`'s explicit design is that it doesn't depend on any of the platform-specific libraries, and this violates that. Is there any reason we can't use a `*mut c_void` like everywhere else?

src/lib.rs
FloVanGH: done!

src/redox.rs
FloVanGH: Thank you for your feedback. Is possible and done.

Osspial: Also, does Redox have a stable ABI, or is using this going to cause safety issues for people that have mixed versions of the redox client toolkit in their project?

FloVanGH: Because Redox is a young project changes are possible. OrbClient itself will stay! If it would help I could be official contact person to Redox specific issues. And I will do my best to help and inform people by changes.

FloVanGH: @Osspial is it ok?

FloVanGH:
Updates?

Lokathor (approved):
Hello @FloVanGH, sorry about the delay! This looks good.

It's been a year, is this still fit to merge or do you want to make any changes?

FloVanGH:
@Lokathor thank you for your answer. Yes it is still fit to merge.

PR #48, opened 2020-01-13
Remove _non_exhaustive fields to use #[non_exhaustive] macro

a1phyr:
Remove `_non_exhaustive` fields and replace them with the new `#[non_exhaustive]` macro

### Pros 
- Disallow users to use this field anyway, which limit future breaking changes
- Cleaner

### Cons
- Rust version support is reduced to 1.40+ (I didn't see any notice about that so it may be fine)
- Breaking change
- Disallow `let handle = Handle { /* fields */ ..Handle::empty() };` syntax

Maybe provide a `Handle::new(...)` function for each type ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
 * **Breaking:** Change type for pointers, which may be null, to `Option<NonNull<_>>`.
+* **Breaking** Remove `_do_not_use` tags to use `#[non_exhaustive]` macro
 
 # 0.3.3 (2019-12-1)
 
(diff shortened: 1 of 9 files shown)


Lokathor:
I like the idea of this, particularly because by now we're about on to 1.50, but I'll need to sit and think about it to be sure.

Lokathor:
Alright, I'm interested in having this in for sure.

@a1phyr there are some merge conflicts it seems. Can you update your PR to fix those? Or if you're not interested any more I can sort it out later.

a1phyr:
Rebased, added Redox, and fixed doc-tests

Lokathor (approved):

Issue #41, opened 2019-12-01
Additional Licenses


Lokathor:
It's traditional for core Rust projects to be MIT OR Apache-2.0. Personally I also like to see the Zlib license as an option as well.

Could we get additional license options for Apache-2.0 at least, and possibly Zlib as well?

This would need a sign off from all those who have added code (I'll list them here _without_ pings and we can ping them if people generally agree with this issue at all).

* Osspial (lead)
* LeshaInc marked stuff as "derive Hash" but I don't think that's significant.
* aloucks similarly put the initial derives on the enum
* dxenonb added readme stuff but not code
* msiglreith renamed a field and added some docs lines
* repi helped fix the last license thing actually, but no code changes.
* ryanisaacg added the web handle
* Didgy74 added XCB and Android.

Osspial:
AFAIK the main reason for Rust projects to be dual-licensed under the MIT and Apache licenses is that the Apache license isn't compatible with the GPL-2.0. I see no reason to go through the effort of multi-licensing when the MIT license is compatible with basically everything.

solarane:
+1. Esp. zlib would definitely be a useful addition.

PR #50, opened 2020-07-30
Add Apache-2.0 and Zlib license options.

Lokathor:
This PR adds the "rust standard" Apache-2.0 license as a license option, as well as the Zlib license because it's even more permissive than the MIT license. The MIT license is of course also still an option.

Being a change in license, this will need the following folks to agree:
* [x] Osspial
* [x] Didgy74
* [x] ryanisaacg
* [x] repi
* [x] msiglreith
* [x] dxenonb
* [x] aloucks
* [ ] LeshaInc
* [x] Dispersia

Closes https://github.com/rust-windowing/raw-window-handle/issues/41


Lokathor:
And a ping to:
@Osspial
@Didgy74
@ryanisaacg
@repi
@msiglreith
@dxenonb
@aloucks
@LeshaInc
@Dispersia

msiglreith:
Thumbs up enough :D?
otherwise,

I license past and future contributions under the 'MIT OR Apache-2.0 OR Zlib' license, allowing licensees to chose either at their option.

aloucks:
I agree.

dxenonb:
I approve of the change.

Didgy74:
I approve.

Lokathor:
@Osspial On the grounds that the only outstanding vote is from @LeshaInc, who seems to have disappeared from github without a way to email them a follow-up, and also since they only added the `Hash` derives, I think that you should now merge this.

PR #51, opened 2020-07-30
Remove the dependency on libc

Lokathor:
This PR completely removes the dependency on `libc`.

It was used to get a `c_void` type, but core::ffi::c_void is sufficient.

It was also used to get `c_ulong`, but we can simply define the type ourselves in the one location that requires it.

Closes https://github.com/rust-windowing/raw-window-handle/issues/3

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,9 +14,6 @@ documentation = "https://docs.rs/raw-window-handle"
 travis-ci = { repository = "rust-windowing/raw-window-handle" }
 appveyor = { repository = "rust-windowing/raw-window-handle" }
 
-[dependencies]
-libc = {version="0.2", features=[]}
-
 [features]
 nightly-docs = []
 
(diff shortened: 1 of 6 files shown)


Osspial:
I'm not particularly happy with having this crate export its own `c_ulong` type. On top of that, merging this PR would require us to release a breaking release, and I don't think dropping the `libc` dependency is enough of an improvement to justify that, especially when most of the higher-level windowing libraries that depend on this pull in `libc` anyway.

Lokathor:
Well it's not as if you'll never release another breaking release ever, because 1.0 is a breaking release, at the minimum.

Also, I myself don't use libc when i use this crate, in fact I avoid this crate partly because of the dumb libc dependency.

Osspial:
If you can figure out a good solution to the `c_ulong` issue, I'd be fine with holding this PR open for whenever the next breaking release happens. Also, we could do a `1.0` release without having it break anything with the [semver trick](https://github.com/dtolnay/semver-trick).

Osspial:
So, it turns out that this change isn't actually breaking, since `libc`'s `c_void` type is just a re-export of `core`'s `c_void` type. I'll go ahead and publish this, and if it turns out to break code we can just yank the release.

Issue #49, opened 2020-05-11
Implement HasRawWindowHandle for RawWindowHandle


kvark:
Currently, there is only one way to go from `winit::Window` to `RawWindowHandle`. Once you get a raw window handle, everything at the lower level has to work with it directly. This is an unfortunate limitation.

For example, [gfx-backend-vulkan::Instance::create_surface](https://docs.rs/gfx-backend-vulkan/0.5.6/gfx_backend_vulkan/struct.Instance.html#method.create_surface) expects `HasRawWindowHandle`. That means we can't get `RawWindowHandle` at any layer above it, but sometimes we really want that, e.g. when we construct `RawWindowHandle` manually on the client side.

msiglreith:
Similar issue: https://github.com/norse-rs/ash-window/pull/10

Osspial:
This was actually implemented at one point (see #29), but it turned out to be unsound so it had to be reverted (see #35). I'd be down with adding a read-only, unsafe-to-construct `RawWindowHandle` wrapper that implements `HasRawWindowHandle` to circumvent that, though.

PR #52, opened 2020-07-30
add TrustedHandle type

Lokathor:
Add a wrapper to mark when a specific raw window handle is to be trusted.

Closes https://github.com/rust-windowing/raw-window-handle/issues/49

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -215,7 +215,7 @@ impl TrustedHandle {
         Self { raw }
     }
 }
-unsafe impl HasRawWindowHandle {
+unsafe impl HasRawWindowHandle for TrustedHandle {
     fn raw_window_handle(&self) -> RawWindowHandle {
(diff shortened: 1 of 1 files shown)


DJMcNab:
Would it be possible to have a safe impl of `new` or equivalent for any `T: HasRawWindowHandle`?

Specifically, an implementation which would replace https://github.com/bevyengine/bevy/blob/a68c78cef5fb2112617de095efd4e64cba1587eb/crates/bevy_winit/src/winit_windows.rs#L117-L119


(I need this for https://github.com/bevyengine/bevy/pull/1279, which is currently assuming that the impl could be correct, so I would like to know if that assumption is wrong anyway. That PR is otherwise blocked on #58 anyway)

Lokathor:
Could you open this as a new issue where I won't lose track of it?

Issue #54, opened 2020-08-13
Android interactions


msiglreith:
Recently, I played a bit with Android in combination with winit/Vulkan/wgpu and there are some topics which should be addressed imo:

1. `raw-window-handle` docs:
> The exact handles returned by raw_window_handle must remain consistent between multiple calls to raw_window_handle, and must be valid for at least the lifetime of the HasRawWindowHandle implementer.

This constraint is in particular hard to uphold for winit as the window handle is only available between certain events in the event loop. As far as I know the ANativeWindow may change ov ertime (Suspend-Resumed cycles).

2. Change the type of `a_native_window` from `*mut _` to `Option<NonNull<_>>` or it should be explictly annotated that it may be `null`. Right now this is not clear, some graphic libraries assume it's always a valid handle. On the other hand, winit currently panicks when requesting the RawWindowHandle as it shouldn't return a nullptr, which basically shifts the issue to the client to uphold the requirement of only requesting the handle if it safe to do so. This makes it actually quite hard to write portable implementations for anything Vulkan based (extensions query and surface support via RawWindowHandle). Using the `Option<NonNull<_>>` would make it safe for libraries consuming the handle as well as window libraries to say that a handle might not be available right now.

msiglreith:
Followup on the 2nd point:

This would probably lead to switch all pointers to either `Option<NonNull<_>>` or `NonNull<_>` to enforce that providers set the corresponding values and make it more clear for clients what to expect - e.g https://github.com/gfx-rs/gfx/issues/3329

Lokathor:
I'm open to adjusting the docs of Android has a different concept of "a window".

PR #55, opened 2020-08-22
Explicitly mark fields which may be null

msiglreith:
- Encode nullable fields of the handle types in the typesystem (`Option<NonNull<_>>`). The current change is not exhaustive as I'm not familiar with some platforms (e.g Web)
- Adjust docs to allow platform cases where the handles may need to be invalidated over time or further gated by platform specifics like winit/android.

Fixes https://github.com/rust-windowing/raw-window-handle/issues/54

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+* **Breaking:** Change type for pointers, which may be null, to `Option<NonNull<_>>`.
+
 # 0.3.3 (2019-12-1)
 
 * Add missing `Hash` implementation for `AndroidHandle`.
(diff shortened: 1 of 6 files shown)


Lokathor (commented):
Windows `HWND` values are just `*mut c_void`, so they can be null. That's generally the "invalid handle" value, and all your OS calls will generally fail and such, but it's a *legal* bit pattern.

Other than that this seems good.

Lokathor (approved):

Lokathor:
Looks good. All set on your end @msiglreith ?

msiglreith:
fine for me 👍

PR #62, opened 2021-01-03
Allow `ios` handles to be used under target_os = "tvos"

thomcc:
I think in an ideal world the handle would be `uikit::UIKitHandle` (not `ios::IOSHandle`) and the variant would be `RawWindowHandle::UIKit` (not `RawWindowHandle::IOS`). The 2nd of these can't be added backwards compatibly (short of using [semver-trick](https://github.com/dtolnay/semver-trick) shenanigans).

This probably holds true for `macos` (should be `appkit`) since the catalyst targets (`*-ios-macabi`, which are `target_os = "ios"`) can use AppKit (in addition to UIKit), although I doubt any Rust code does this.

All that said, the renaming suggestions are just food for thought for when 1.0.0 happens tho, since it will probably need to use the semver-trick anyway to avoid ecosystem fragmentation, but is a opportunity to fix imperfect naming choices like this.

### DIFF
--- a/src/ios.rs
+++ b/src/ios.rs
@@ -1,7 +1,8 @@
 use core::ffi::c_void;
 use core::ptr;
 
-/// Raw window handle for iOS.
+/// Raw window handle for IOS and other platforms using UIKit-based windowing
+/// (tvOS, Mac Catalyst (`*-ios-macabi` targets), and, in theory, watchOS).
 ///
 /// ## Construction
 /// ```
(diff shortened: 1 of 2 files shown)


Lokathor:
It seems like https://github.com/rust-windowing/raw-window-handle/pull/64 supersedes this, right?

Lokathor:
Based on a Discussion with Thomcc on discord, they won't have time to continue this PR, so i'm going to close it in favor of a broader change like in #64

Issue #60, opened 2020-12-21
Use separate types for each platform


chrisduerr:
I find the current API a bit strange, since it basically offers an enum that just has a single member on most platforms.

Is there any reason why the `RawWindowHandle` isn't just a platform-specific `type` that is decided at compile-time? It's not like people are going to be matching against the platform's implementation at runtime, especially since this is impossible because the enum has only the members available for the current platform.

grovesNL:
The variants don't all match 1:1 with a platform, so the variant isn't known at compile-time. For example, [Xlib, Xcb, and Wayland are all available on `linux`](https://github.com/rust-windowing/raw-window-handle/blob/master/src/lib.rs#L108-L172)

chrisduerr:
Yeah, which is not a problem. You can still provide an enum for platforms that actually make a decision at runtime.

But adding a runtime check for something that can be uniquely identified at compile time just leads to unergonomic code.

Lokathor:
I'm inclined to go in the direction of https://github.com/rust-windowing/raw-window-handle/issues/63. We keep the enum, and just remove all the cfg stuff.

Issue #58, opened 2020-10-25
New release? 


kettle11:
It's been almost a year since the last release and I've been looking forward to using the version of `raw-window-handle` that doesn't depend on `libc`. Is there any chance of a new release soon?

msiglreith:
Hi, I feel that there a bunch of open topics for this library, but currently missing a maintainer to decide and take care of it.
As this is such a core piece to the ecosystem more hands involved would be awesome - we rely too much on Osspial doing great stuff ❤️

Lokathor:
I just took over, but I'll try to get a new release "very soon".

I haven't looked too closely at the library since like 0.1 or 0.2, so I want to orient myself a bit first.

PR #67, opened 2021-01-25
Get ready for 0.4

Lokathor:
We've got some breaking changes building up and it's time to move to a 0.4

* [ ] TODO: There's some docs in the related PR from Thomcc that we should incorporate here.
* [ ] TODO: we need to rename ios to uikit and macos to appkit, since you can use the macos types on target_os="ios" (via catalyst) and the ios types on target_os!="ios" (tvos, for example)

---
* Closes https://github.com/rust-windowing/raw-window-handle/issues/63
* Closes https://github.com/rust-windowing/raw-window-handle/issues/60
* Closes https://github.com/rust-windowing/raw-window-handle/issues/58

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * **Breaking:** Change type for pointers, which may be null, to `Option<NonNull<_>>`.
 * **Breaking** Remove `_do_not_use` tags to use `#[non_exhaustive]` macro
+* Added `TrustedWindowHandle::from_has_raw_window_handle`.
 
 # 0.3.3 (2019-12-1)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,43 +1,46 @@
-# Unreleased
+# Changelog
+
+## 0.4.0 (?)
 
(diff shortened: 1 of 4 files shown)


DJMcNab:
Maybe a better name for `TrustedRawWindowHandle` would be something like `StableRawWindowHandle` or similar, since the guarantee is that the inner handle won't be mutated.

i509VCB:
I'm curious here, I've noticed some of the handles like macOS, Windows, Android and iOS use `Option<NonNull<c_void>>` while Wayland, XCB and Xlib still use `*mut c_void`. Is this inconsistency intentional or were those specific handles not changed yet to be in line with the other handles.

Lokathor:
I think it was an accidental oversight.

However, I also think that it's a bad change in retrospect. Every pointer here can optionally be null, that's just a blanket rule. I think actually using Option<NonNull<c_void>> makes things less ergonomic though. Realistically, we'll probably want to undo that change before the next release.

Lokathor:
Merging this partial progress, and the rest will be finished up in https://github.com/rust-windowing/raw-window-handle/pull/70

Issue #72, opened 2021-11-13
Publish version 0.4 on crates.io


leafjolt:
Since #70 was merged,  I was wondering if there is anything left to do before releasing version 0.4 on crates.io, or if the release can go forward.

maroider:
I think we're still blocked on reaching consensus in #71.

leafjolt:
Would it be possible to release a new version that leaves out the safety changes/TrustedWindowHandle for now while #71 is sorted out but includes the changes to platform elements, such as removing cfg guards, changing `IOS` to `UiKit` (and other similar changes) and adding Redox/WinRt support? I think that the platform changes are still somewhat significant and would be useful to many users.

maroider:
I believe it is feasible to do so.

cc @Lokathor

Lokathor:
If no one's opposed I can probably get to this on Monday or Tuesday.

Lokathor:
published! https://docs.rs/raw-window-handle/0.4.0/raw_window_handle/index.html

maroider:
Errrm, did you miss this part?

> Would it be possible to release a new version that **leaves out the safety changes/TrustedWindowHandle for now while #71 is sorted out**

Lokathor:
i thought that those were *already* "out" and so i could just publish the `main` branch.

Lokathor:
Okay I yanked 0.4.0, sorry for the confusion.

So... what would a PR that makes the crate better for wgpu look like. I've never used wgpu ever so maybe I don't get what's "going stale". @maroider can you do a PR for whatever you need changed.

madsmtm:
I think this can be closed.

On a related note, I've opened #74 to alleviate the need for ecosystem-wide upgrade.

Issue #71, opened 2021-06-16
Improve safety around `HasRawWindowHandle`


maroider:
We got a bit [sidetracked](https://github.com/rust-windowing/raw-window-handle/pull/70#issuecomment-855274656) by this in #70, so I'm opening this issue to track it separately.

To summarize the problem briefly:
**Letting users use windows with crates like `wgpu` without requiring the user to write any unsafe code is desirable.**
`TrustedWindowHandle` can't really be used safely after it has been around for a bit. Unfortunately, there's no guarantee that the underlying window handle won't become invalid while a `TrustedWindowHandle` is alive.
gfx-rs/wgpu#1463 has a more in-depth outline of the issue at hand, and suggests that `raw-window-handle` should add the following impls of `HasRawWindowHandle`:
```rust
impl<'a, T: HasRawWindowHandle> HasRawWindowHandle for &'a T { /* .. */ }
impl<T: HasRawWindowHandle> HasRawWindowHandle for Rc<T> { /* .. */ }
impl<T: HasRawWindowHandle> HasRawWindowHandle for Arc<T> { /* .. */ }
```

The addition of these impls of `HasRawWindowHandle` makes `TrustedWindowHandle` seem kind of useless (for lack of a less biting word).

Lokathor:
At a glance, I don't understand what the extra HasRawWindowHandle impls accomplish, because don't any of those values just deref to the contained type when necessary anyway?

maroider:
It's not explicitly documented as a guarantee of `HasRawWindowHandle`, but `winit` and `sdl2`'s window structs destroy the underlying window when they're dropped.
`wgpu` wants to make sure that a `wgpu::Surface` is always valid, and as such they need to make sure the `Window` isn't destroyed while the `Surface` is alive, because there's no guarantee in the spec that a Vulkan surface will keep the underlying window handle valid.
`Surface` *could* just expose a method that allows you to access the underlying window, but that's also not the most convenient thing in the world.
The extra impls allow for convenient access to the underlying window type without having to go through such accessors. Glium is a particularly bad example here (even though Glium doesn't use `raw-window-handle`) in that you need to go through two layers of accessors to get at the `winit::window::Window`.

maroider:
And I forgot to do earlier, but cc @pythonesque @kvark

Lokathor:
Hm, well those are all reasonable impls to have, so I'm for it if that fixes things for wgpu.

pythonesque:
[...]

PR #73, opened 2021-11-15
Remove `TrustedWindowHandle`

maroider:
As per @Lokathor's [request](https://github.com/rust-windowing/raw-window-handle/issues/72#issuecomment-969289187) in #72, I have made some changes that I believe need to be made.

1. I removed `TrustedWindowHandle` as its usefulness is somewhat dubious.
2. I added extra impls for `HasRawWindowHandle` as proposed in #71.
    I believe these impls represent a decent enough usability improvement to be added as-is.

cc @pythonesque

###### PS: I missed a `0.` in the branch name, lol


src/lib.rs
```diff
@@ -18,6 +18,8 @@
 //! be used along with the struct update syntax to construct it. See each specific struct for
 //! examples.
 
+extern crate alloc;
```
Lokathor, line 21: This is good, but the use of the alloc crate and the extra impls should be behind an `alloc` feature.

src/lib.rs
```diff
@@ -52,6 +54,22 @@ pub unsafe trait HasRawWindowHandle {
...
+    fn raw_window_handle(&self) -> RawWindowHandle {
+        (**self).raw_window_handle()
+    }
+}
+unsafe impl<T: HasRawWindowHandle> HasRawWindowHandle for alloc::sync::Arc<T> {
+    fn raw_window_handle(&self) -> RawWindowHandle {
+        (**self).raw_window_handle()
+    }
+}
+
```
madsmtm, line 72: I agree with adding these impls since they are consistent with the current wording of the safety guarantees implementors of `HasRawWindowHandle` must uphold.

If we find in https://github.com/rust-windowing/raw-window-handle/issues/71 or likewise that the safety guarantees should change, we can then consider if these impls should change as well.

madsmtm (approved):

src/lib.rs
maroider: Yeah, that's my general reasoning behind pushing this through for now.

maroider:
The impls for `Rc<T>` and `Arc<T>` should be behind the new `alloc` feature now.

Lokathor (approved):

Lokathor:
This looks good. I'll merge it and publish later today.

Publishing as 0.4.1 should be fine, since the averted 0.4.0 was only out for like 20 minutes.

Lokathor:
https://docs.rs/raw-window-handle/0.4.1/raw_window_handle/

PR #74, opened 2021-11-19
Alleviate the need for ecosystem-wide upgrading to 0.4.1

madsmtm:
~This PR is opened against `master`, but in reality it should be merged into 5207e99a5872e06de2807f556222d056703af44a.~

Using the [semver trick](https://github.com/dtolnay/semver-trick), we can make upgrading from `0.3` to `0.4.1` much less painful!
We simply add a `v0.3` `HasRawWindowHandle` impl for a `v0.4.1` `HasRawWindowHandle`, and release `v0.3.4`.

The idea is that this will allow crates whose purpose is to provide an `HasRawWindowHandle` implementation to upgrade first (e.g. `winit`, `sdl2`, `glfw`, `fltk`, ...). When a sufficient amount of "provider" crates have upgraded, "consumer" crates (e.g. `gfx`, `wgpu`, `rfd`, ...) can safely start upgrading while minimizing breakage for downstream users.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ appveyor = { repository = "rust-windowing/raw-window-handle" }
 
 [dependencies]
 libc = {version="0.2", features=[]}
+new = { version = "0.4.1", package = "raw-window-handle" }
 
 [features]
 nightly-docs = []
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Unreleased
 
+* Add `HasRawWindowHandle` implementation for `HasRawWindowHandle` in the newer
+  version `0.4.1`.
+
+  This allows "provider" crates that implement `HasRawWindowHandle` (like
+  `winit`, `sdl2`, `glfw`, `fltk`, ...) to upgrade to `v0.4.1` without a
+  breaking change.
+
+  Afterwards "consumer" crates (like `gfx`, `wgpu`, `rfd`, ...) can start
+  upgrading with minimal breakage for their users.
+
 # 0.3.3 (2019-12-1)
 
 * Add missing `Hash` implementation for `AndroidHandle`.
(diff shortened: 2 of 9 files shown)


madsmtm:
It would be helpful if someone could create a `0.3` branch targetting 5207e99a5872e06de2807f556222d056703af44a, then it becomes much easier to review and discuss this PR

maroider:
I hope this was what you wanted :)

Cargo.toml
```diff
@@ -16,6 +16,7 @@ appveyor = { repository = "rust-windowing/raw-window-handle" }
 
 [dependencies]
 libc = {version="0.2", features=[]}
+new = { version = "0.4.1", package = "raw-window-handle" }
```
madsmtm, line 19: Not sure if this, or by using `extern crate raw_window_handle as new;` is the most robust way to rename a crate?

madsmtm:
If you agree that this is desirable, I'm thinking the way forward is to:
- We merge this
- Add a second PR (or just a commit) that releases `v0.3.4` (and updates the changelog)
- Finally merge the `semver-trick` branch into `master` again (and in the merge remove everything that this PR added, except the changelog entry)

Lokathor (approved):

Lokathor:
This all seems like a good idea, but i've never done the semver trick thing before.

First, I should just merge this, right?

madsmtm:
Well, I haven't done it before either, but I'm fairly certain it will work, so yes, merge this and make a release `v0.3.4`, I can also open a PR for that (but since the changelog entry should include the release date it seems a bit brittle).

It should be noted that adding blanket implementations of `HasRawWindowHandle` is technically a breaking change in case a downstream crate manually implements both the old and new version of the trait. This is very, very unlikely, given how little time `0.4` has been released for.

Lokathor:
Okay, i'll merge this now then.

PR #75, opened 2021-11-19
Allow unsized handle types for `Arc` and `Rc`

a1phyr:
Lower bounds of `HasRawWindowHandle` impls of `Arc` and `Rc` to allow unsized types.

This allows to use eg `Arc<dyn HasRawWindowHandle + Send + Sync>`, which is nice to avoid generics.

However this is not done on references because it is a fundamental type, so lowering the bound is a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # Changelog
 
-## 0.4.1 (?)
+## 0.4.2 (?)
+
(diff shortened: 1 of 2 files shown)


Lokathor:
if it weren't a breaking change would we also want this on references? It seems like yes.

a1phyr:
I think that we eventually want it on references. In addition to the above advantage about avoiding generics, it costs nearly nothing, doesn't make the crate harder to maintain but makes it  more consistent.

Lokathor (approved):

PR #81, opened 2021-11-27
Grab the changelog entry for `v0.3.4`

madsmtm:
Not sure how you want to merge this (my `git` skills are far from superb), but the resulting commit should consist of only the new changelog entry.

PR #82, opened 2021-11-29
Update CHANGELOG.md

Lokathor:
@madsmtm i think this PR does what you were trying to do in https://github.com/rust-windowing/raw-window-handle/pull/81, if this looks right to you we can merge it. The other one had a bunch of other changes bundled into the diff that looked a little strange so I think git was confused somehow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,5 @@
 # Changelog
 
-# 0.3.4 (2021-11-27)
-
-* Add `HasRawWindowHandle` implementation for `HasRawWindowHandle` in the
-  newer `v0.4`.
-  This allows "provider" crates that implement `HasRawWindowHandle` (like
-  `winit`, `sdl2`, `glfw`, `fltk`, ...) to upgrade to `v0.4` without a
-  breaking change.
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,15 @@
 # Changelog
 
+# 0.3.4 (2021-11-27)
```
madsmtm, line 3: Nit: I would place this between `0.3.3` and `0.4.0`, since users that have already upgraded to `0.4.2` don't really care about this (while it is relevant to `0.3.3` users, and to users about to upgrade to the `0.4`-line)

madsmtm (approved):
This looks fine, with #81 I was hoping to somehow get rid of the `semver-trick` branch while preserving history, but I guess that's not really possible :/

maroider:
Does #81 even have a commit where you remove the non-changelog bits?

madsmtm:
> Does #81 even have a commit where you remove the non-changelog bits?

I would have done it with a standard merge commit, and when merging, I'd rip out the non-changelog bits

> we only use "squash and merge", so the git history would still be lost.

Totally fair

PR #83, opened 2021-12-02
Add visual IDs to X11 handles

Ralith:
Needed for Vulkan physical device presentation support queries (e.g. [`vkGetPhysicalDeviceXlibPresentationSupportKHR`](https://khronos.org/registry/vulkan/specs/1.2-extensions/man/html/vkGetPhysicalDeviceXlibPresentationSupportKHR.html)).

I'm not sure what, if anything, would be a sensible default value for these that won't cause fallout if someone tries to use it. Do we try to solve that problem in general?

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -18,6 +18,8 @@ pub struct XlibHandle {
     pub window: c_ulong,
     /// A pointer to an Xlib `Display`.
     pub display: *mut c_void,
+    /// An Xlib visual ID, or 0 if unknown.
+    pub visual_id: c_ulong,
 }
 
 /// Raw window handle for Xcb.
@@ -35,6 +37,8 @@ pub struct XcbHandle {
     pub window: u32, // Based on xproto.h
     /// A pointer to an X server `xcb_connection_t`.
     pub connection: *mut c_void,
+    /// An X11 `xcb_visualid_t`, or 0 if unknown.
+    pub visual_id: u32,
 }
 
 /// Raw window handle for Wayland.
@@ -59,6 +63,7 @@ impl XlibHandle {
         Self {
             window: 0,
             display: ptr::null_mut(),
+            visual_id: 0,
         }
     }
 }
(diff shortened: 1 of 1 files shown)


Lokathor:
If 0 is generally "invalid" then that's fine as the default.

Ralith:
That doesn't seem to be documented either way, though empirically it seems to hold and https://github.com/mirror/libX11/blob/5faa8dc0b364f55f19034c370436074646d9ad54/src/GetRGBCMap.c#L54 suggests it's sensible.

madsmtm:
You could perhaps submit a patch to their documentation, saying that a visual ID that is `None` / zero is indeed invalid? E.g. add it to the [Visual Types](https://github.com/mirror/libX11/blob/master/specs/libX11/CH03.xml) part of their specification.

And [here](https://github.com/mirror/libX11/blob/5faa8dc0b364f55f19034c370436074646d9ad54/specs/libX11/CH14.xml#L5212-L5214), mentioning that `If the visualid is not present` means `the visualid is None`.

Ralith:
I don't think I have the energy to work out the contribution process for Xlib.

> And here, mentioning that If the visualid is not present means the visualid is None.

Where is that meaning specified?

https://xcb.freedesktop.org/xlibtoxcbtranslationguide/ does also seem to use 0 as a sentinel.

Lokathor:
@Ralith how about just expanding the doc comments a bit on our end. Just add a line that says "the current default is 0, which seems to be the standard sentinel for a missing value of this type" or something like that.

madsmtm:
> I don't think I have the energy to work out the contribution process for Xlib.

Yeah, me neither, I was hoping you might already be involved.

> Where is that meaning specified?

Sorry, I meant you (or a person contributing to the documentation) would update the docs there too, to specify what "not present" means.

> expanding the doc comments a bit on our end

Fine by me.

Ralith:
How's this?

Lokathor (approved):

Lokathor:
excellent.

PR #88, opened 2022-01-22
add haiku support

hoanga:
this pr adds haiku support (BWindow and BDirectWindow)

the following is an example run of tests on a haiku instance:

```
> uname -a
Haiku shredder 1 hrev55602 Oct 29 2021 07:10:38 x86_64 x86_64 Haiku

> cargo test
   Compiling cty v0.2.1
   Compiling raw-window-handle v0.4.2 (/boot/home/src/git/rust-libs/raw-window-handle)
    Finished test [unoptimized + debuginfo] target(s) in 1.76s
     Running unittests (target/debug/deps/raw_window_handle-c16f3ff9cf02e130)

running 0 tests

test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s

   Doc-tests raw-window-handle

running 11 tests
test src/android.rs - android::AndroidNdkHandle (line 7) ... ok
test src/haiku.rs - haiku::HaikuHandle (line 7) ... ok
test src/redox.rs - redox::OrbitalHandle (line 7) ... ok
test src/appkit.rs - appkit::AppKitHandle (line 7) ... ok
test src/windows.rs - windows::WinRtHandle (line 33) ... ok
test src/unix.rs - unix::XlibHandle (line 9) ... ok
test src/windows.rs - windows::Win32Handle (line 7) ... ok
test src/web.rs - web::WebHandle (line 4) ... ok
test src/uikit.rs - uikit::UiKitHandle (line 7) ... ok
test src/unix.rs - unix::XcbHandle (line 28) ... ok
test src/unix.rs - unix::WaylandHandle (line 47) ... ok

test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.08s

```

### DIFF
--- a/src/haiku.rs
+++ b/src/haiku.rs
@@ -1,0 +1,28 @@
+use core::ffi::c_void;
+use core::ptr;
+
+/// Raw window handle for Haiku.
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::HaikuHandle;
+/// let mut handle = HaikuHandle::empty();
+/// /* set fields */
+/// ```
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct HaikuHandle {
+    /// A pointer to a BWindow object
+    pub b_window: *mut c_void,
+    /// A pointer to a BDirectWindow object
+    pub b_direct_window: *mut c_void,
+}
+
+impl HaikuHandle {
+    pub fn empty() -> Self {
+        Self {
+            b_window: ptr::null_mut(),
+            b_direct_window: ptr::null_mut(),
+        }
+    }
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,6 +30,7 @@ mod uikit;
 mod unix;
 mod web;
 mod windows;
+mod haiku;
 
 pub use android::AndroidNdkHandle;
 pub use appkit::AppKitHandle;
@@ -38,6 +39,7 @@ pub use uikit::UiKitHandle;
 pub use unix::{WaylandHandle, XcbHandle, XlibHandle};
 pub use web::WebHandle;
 pub use windows::{Win32Handle, WinRtHandle};
+pub use haiku::HaikuHandle;
 
 /// Window that wraps around a raw window handle.
 ///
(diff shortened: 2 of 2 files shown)


Lokathor:
Well I don't know anything about Haiku, but other than that things look good.

Is there any sort of official documentation that should maybe be linked to for these fields?

maroider:
Maybe run `cargo fmt`? I'm pretty sure `rustfmt` would've sorted the module declarations.

hoanga:
closest that i know of documentation that looks official is the api documentation [here](https://www.haiku-os.org/docs/api/index.html).  that has (a little) more information regarding [BWindow](https://www.haiku-os.org/docs/api/classBWindow.html) and [BDirectWindow](https://www.haiku-os.org/docs/api/classBDirectWindow.html)

i have also pushed an update with cargo fmt applied.

maroider:
If I'm reading these docs correctly, then `BDirectWindow` is a subclass of `BWindow`, and you can't create a `BDirectWindow` from a `BWindow`, so you have to decide if you want a `BDirectWindow` when creating the window? If that's the case, then it might make sense to document that `b_direct_window` *may* be null. I also saw some references to a `BWindowScreen`, but I couldn't find a page for it in the documentation.

hoanga:
i have added a note about the possibility for null for BDirectWindow. i am not sure of a case where one would want to create a BDirectWindow from BWindow. at the time of window creation, the choice would stem (at least according to the documentation) from whether the caller wishes to use the app_server or attempt to bypass it. for the former case, use BWindow. in the other case, a BDirectWindow.

in regards to BWindowScreen i also noticed some note(s) on it but no real api documentation outside of some brief mentions in some documents that appeared to be in a legacy (aka outdated) area of documentation so left it out.

for BDirectWindow, from what i can tell it seems that the current direction is to prefer BWindow over BDirectWindow. if it would be preferred to only include what is necessary, i am fine with leaving out the BDirectWindow reference and revisiting that at another time. however, libraries ported to haiku that deal with windowing such as libsdl had been using BDirectWindow so opted to keep it in (c.f. [this](https://github.com/libsdl-org/SDL/pull/5218))

Lokathor (approved):

Issue #35, opened 2019-10-24
RawWindowHandle's HasRawWindowHandle implementation is unsound


Osspial:
The `impl HasRawWindowHandle for RawWindowHandle` added in #29 is a soundness hole, since it allows safe code to instantiate an invalid handle and pass it to a function that relies on `HasRawWindowHandle` being implemented correctly:

```rust
fn use_raw_window_handle<H: HasRawWindowHandle>(h: H) {/*impl*/}

use_raw_window_handle(RawWindowHandle::Windows(WindowsHandle {
    hwnd: 41 as *mut c_void,
    hinstance: 43 as *mut c_void,
    ..WindowsHandle::empty()
}));
```

Removing that implementation is trivial, but there are two unresolved questions:

- Should we expose some sort of `pub RawWindowHandleActive(RawWindowHandle)` struct that can only be initialized with an unsafe function?
- Does removing the impl require a bump to `0.4.0`? The official Rust semver guidelines say that soundness fixes don't count as breaking changes, so I'm leaning towards no.

icefoxen:
Far, far from an expert but if you provide a trait that provides soundness guarantees for other people to implement and document how to impl it correctly, then it's their problem if they don't impl it correctly. It's your job to *try* to work with both upstream and downstream to make things work but you can't force either to comply.

Lokathor:
No, the fault is that an implementing struct in this crate does it wrong

just delete the struct. Few if any people will ever need to add this trait to their type, and they're pro enough to be able to at least read some docs.

Ralith:
Is this hypothetical safety guarantee actually a practically one to begin with? What do graphics APIs do if you delete the window out from under them after setup? My intuition here is that anything working with raw window handles is necessarily unsafe.

Ralith:
I think releasing 0.3.1 and yanking 0.3.0 would be acceptable if deemed necessary, but I'm not sure there are meaningful safety gains to be had, presuming that most things that take a `HasRawWindowHandle` will ultimately capture the resulting handle without lifetime guarantees.

Lokathor:
The issue is that the contract so far is that by implementing the trait you're assuring that whatever you return is valid and that unsafe code can trust you since you're also unsafe code.

And then this breaks that because unsafe code shouldn't generally trust safe code.

Ralith:
[...]

PR #89, opened 2022-02-17
Implement `HasRawWindowHandle` for `RawWindowHandle`

tasgon:
Just a small QoL change


maroider:
This is unsound, as `RawWindowHandle` can't keep the underlying window alive, as required by any impl of `HasRawWindowHandle` (see #35,  #49, #71, and #73).

PR #93, opened 2022-07-06
Decouple display and window from `RawWindowHandle`

kchibisov:
On free platforms there're usually multiple displays, so APIs like
EGL have a concept of native display when creating EGLDisplay.

Sometimes creating EGLDisplay must be done before creating the window
on that particular display, to supply e.g. X11 visual information to
the window that will be created on that display later on.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,4 +2,8 @@
 [![Crates.io](https://img.shields.io/crates/v/raw-window-handle.svg?maxAge=2592000)](https://crates.io/crates/raw-window-handle)
 [![Docs](https://docs.rs/raw-window-handle/badge.svg)](https://docs.rs/raw-window-handle)
 
-This library provides standard types for accessing a window's platform-specific raw window handle. This does not provide any utilities for creating and managing windows; instead, it provides a common interface that window creation libraries (e.g. Winit, SDL) can use to easily talk with graphics libraries (e.g. gfx-hal).
+This library provides standard types for accessing a window's platform-specific
+raw window handle and display's platform-specific raw display handle. This does
+not provide any utilities for creating and managing windows; instead, it
+provides a common interface that window creation libraries (e.g. Winit, SDL)
+can use to easily talk with graphics libraries (e.g. gfx-hal).
--- a/src/android.rs
+++ b/src/android.rs
@@ -1,22 +1,40 @@
 use core::ffi::c_void;
 use core::ptr;
 
-/// Raw window handle for Android NDK.
+/// Raw display handle for Android.
 ///
 /// ## Construction
 /// ```
-/// # use raw_window_handle::AndroidNdkHandle;
-/// let mut handle = AndroidNdkHandle::empty();
+/// # use raw_window_handle::AndroidDisplayHandle;
+/// let mut display_handle = AndroidDisplayHandle::empty();
 /// /* set fields */
 /// ```
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct AndroidNdkHandle {
+pub struct AndroidDisplayHandle;
+
+impl AndroidDisplayHandle {
+    pub fn empty() -> Self {
+        Self {}
+    }
+}
+
+/// Raw window handle for Android NDK.
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::AndroidNdkWindowHandle;
+/// let mut window_handle = AndroidNdkWindowHandle::empty();
+/// /* set fields */
+/// ```
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct AndroidNdkWindowHandle {
     /// A pointer to an `ANativeWindow`.
     pub a_native_window: *mut c_void,
 }
 
-impl AndroidNdkHandle {
+impl AndroidNdkWindowHandle {
     pub fn empty() -> Self {
         Self {
             a_native_window: ptr::null_mut(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,16 +1,17 @@
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-//! Interoperability library for Rust Windowing applications.
+//! Interoperability library for Rust Windowing applications and Rust applications that connect
+//! to system display server.
 //!
 //! This library provides standard types for accessing a window's platform-specific raw window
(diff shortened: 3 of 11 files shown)


src/android.rs
```diff
@@ -1,22 +1,40 @@
...
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::AndroidDisplayHandle;
+/// let mut display_handle = AndroidDisplayHandle::empty();
+/// /* set fields */
+/// ```
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct AndroidDisplayHandle;
```
MarijnS95, line 14: There's no handle in here, yet should it follow the `AndroidNdkDisplayHanlde` naming scheme?

src/lib.rs
```diff
@@ -101,68 +105,190 @@ pub enum RawWindowHandle {
...
+/// Note that all variants are present on all targets (none are disabled behind
+/// `#[cfg]`s), but see the "Availability Hints" section on each variant for
+/// some hints on where this variant might be expected.
+///
+/// Note that these "Availability Hints" are not normative. That is to say, a
+/// [`HasRawDisplayHandle`] implementor is completely allowed to return something
+/// unexpected. (For example, it's legal for someone to return a
+/// [`RawDisplayHandle::Xlib`] on macOS, it would just be weird, and probably
+/// requires something like XQuartz be used).
+pub enum RawDisplayHandle {
```
MarijnS95, line 227: What was the design decision behind having a variant for every platform, even if the platform doesn't have any display handle?

Also, missing:

```rust
#[non_exhaustive]
```

If this is `non_exhaustive` I see even less reason to provide empty structs up-front, we can always introduce things later.

src/android.rs
kchibisov: I don't think so, the point is that it's a general handle for android, in theory it could be used regardless of NDK.

src/lib.rs
kchibisov: I was copying the behavior of `RawWindowHandle` and it's a `non_exhaustive`.

If the platform doesn't have a display handler(as in there's no value for it) it should still provide an indicator that it's e.g. Windows or Android, since on those platforms it means that connection could be done in some other way. Basically I just need an indicator to know what platform I'm creating display for in EGL.

kchibisov: Added missing derives for `RawDisplayHandle`.

Lokathor:
As I said in the issue for this, I don't have any big expertise here and will try to defer to others on this as much as possible.

However, I want to ask if we could make this a non-breaking update by just adding a new type, but leaving the old type alone? Or would that just be weird? I'd prefer a non-breaking update if possible, but I understand if that makes things confusing because data can be in two separate places or whatever the problem would be.

kchibisov:
>However, I want to ask if we could make this a non-breaking update by just adding a new type, but leaving the old type alone? Or would that just be weird?

It could be weird since you'll have 2 sources of truth. I think it's fine to go with breaking update here, since the changes are sort of trivial and winit will update for them right away (in 0.27 release).

I understand that some libraries will take time for them to update, but I don't see a reason to not break, it's not like this crate is stable...

Though if you insist we could probably mark old field on the structs as deprecated, but that would mean that we can't rename them either.

kchibisov:
@Lokathor if you want me to make a non breaking change, let me explicitly know that, so we can proceed with it and update winit before next release.

msiglreith:
I'm fine with the proposed changes but adding/keeping a combined `WindowHandle` + trait would be good imo.
Besides EGL most APIs use both information on surface creation, which would require API changes in the consuming libraries as well resulting in user code changes.

kchibisov:
Hm, I think I'll leave old structure untouched and add a method for it to retrieve `RawDisplayHandle` as well as a separate trait for display, since that's the thing winit's event loop will implement. Because with e.g. GLX you need to create glx config before creating winit window, since you must pass a visual for it.

Lokathor:
@kchibisov oh sorry for being kinda vague. a break like this is fine.

kchibisov:
@Lokathor so, should I keep everything the way it's right now or we should have a type that accumulates in itself `RawWindowHandle` and `RawDisplayHandle`?

Lokathor:
Separate like it is now is probably fine.

I'm not sure I'll have time for doing a review+release before Tuesday, but if anyone else gives it an approval review I can at least merge the PR into the main branch.

README.md
```diff
@@ -2,4 +2,8 @@
 [![Crates.io](https://img.shields.io/crates/v/raw-window-handle.svg?maxAge=2592000)](https://crates.io/crates/raw-window-handle)
 [![Docs](https://docs.rs/raw-window-handle/badge.svg)](https://docs.rs/raw-window-handle)
 
-This library provides standard types for accessing a window's platform-specific raw window handle. This does not provide any utilities for creating and managing windows; instead, it provides a common interface that window creation libraries (e.g. Winit, SDL) can use to easily talk with graphics libraries (e.g. gfx-hal).
+This library provides standard types for accessing a window's platform-specific
+raw window handle and display's platform-specific raw display handle. This does
+not provide any utilities for creating and managing windows; instead, it
+provides a common interface that window creation libraries (e.g. Winit, SDL)
+can use to easily talk with graphics libraries (e.g. gfx-hal).
```
maroider, line 9:
```suggestion
This library provides standard types for accessing a window's platform-specific
raw window handle, as well as any handles required to talk to the platform's
display server.
```
I'm not 100% happy with my re-formulation here,  but I feel like appending "and display's platform-specific raw display handle" to the first sentence here doesn't quite work. (ps: this suggestion isn't intended to be applied through the web UI)

src/lib.rs
```diff
@@ -1,16 +1,17 @@
 #![no_std]
 #![cfg_attr(docsrs, feature(doc_cfg))]
 
-//! Interoperability library for Rust Windowing applications.
+//! Interoperability library for Rust Windowing applications and Rust applications that connect
+//! to system display server.
```
maroider, line 5:
```suggestion
//! Interoperability library for Rust Windowing applications.
```

Leaving this as-is might be fine, as this is still the primary purpose of `raw-window-handle`.

README.md
Lokathor: I see what you're getting at with the wording being, maybe not the best. However, I think that the idea comes across, and further docs updates for this can be done later with non-breaking releases, so let's continue as is.

Lokathor (approved):

README.md
maroider: Sure

maroider (approved):

PR #95, opened 2022-07-14
Add Generic Buffer Manager handles

kchibisov:


kchibisov:
>I've never heard of these. Can you make the descriptions link to a man page or other documentation about them? If not it's fine, but it would be preferable.

It's a little bit of a niche platform and folks who know it we'll likely now what to look for.

In general this is a mechanism used to allocate buffers for OpenGL, Vulkan, etc, used by mesa.

And you can use it directly if you fell like with e.g. EGL.

I'd add a note about that.

kchibisov:
I've added a note on what to use it with.

Also, @Lokathor would you mind bumping `raw-window-handle` to `0.5.0` after that?

Lokathor (approved):

Lokathor:
I can approve and merge from my phone easily enough, but for a release I need a proper computer. I will try to remember to do a release later today.

PR #96, opened 2022-07-14
Bump version to 0.5.0

kchibisov:
--

Note, I can't push to crates.io, just opening a PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## Unreleased
+## 0.5.0 (2022-07-14)
 
 * **Breaking:** The `RawWindowHandle` variants were split into `RawDisplayHandle` and `RawWindowHandle`.
 * The X11 screen is now present in new `XlibDisplayHandle` and `XcbDisplayHandle`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "raw-window-handle"
-version = "0.4.3"
+version = "0.5.0"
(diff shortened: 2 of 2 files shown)


Lokathor:
yeah this is good.

Usually i use the `cargo-release` tool which also makes a tag in the repo and stuff. i can probably do that later today.

Lokathor (approved):

maroider:
Is it feasible for us to do the "semver trick" again?

kchibisov:
>Is it feasible for us to do the "semver trick" again? It would be nice to avoid ecosystem churn again.

I don't think there's a need to. The point is that we can likely make the crate stable with such api (though, I'd wait for rust to update its core, so we can remove cty dep and have zero dependencies).

Breaking it when we can break is better in my opinion. Since this migration will happen sooner or later, and in case of rust I'd suggest to make it sooner.

Lokathor:
ctypes got stabilized in core::ffi like *this morning*, so we're very close to that

https://github.com/rust-lang/rust/pull/98315

kchibisov:
That's nice, so it won't take long. I'd wait for the rust version with those stable variants to reach major linux distributions and update to it releasing the `v1.0.0` version.

Lokathor:
Well that will hit stable in about 12 weeks.

How far behind "latest stable" are most major Linux variants?

kchibisov:
>How far behind "latest stable" are most major Linux variants?

Like 2 version or more (if we talk about Ubuntu and friend). So we'd have like half a year...

Though, we're not in a rush here, maybe the current design would need adjustments and recent changes would cause issues, so take it as a time before stabilization.

PR #97, opened 2022-08-10
Make implementors of HasRawWindowHandle v0.5 also implement v0.4's

madsmtm:
Similar approach as in https://github.com/rust-windowing/raw-window-handle/pull/74

Will follow up with a changelog entry to `master` afterwards.

Fixes `winit` issue https://github.com/rust-windowing/winit/issues/2415

CC @rib @maroider

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,16 @@
 # Changelog
 
+## 0.4.4 (2022-08-10)
+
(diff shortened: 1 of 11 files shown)


madsmtm:
Ah no, of course we can't do it like this because I'm removing the blanket impls, duh...

PR #100, opened 2022-08-11
Transition to GitHub Actions

madsmtm:
TravisCI is dead, and using GitHub Actions instead of AppVeyor means we use the same setup across all `rust-windowing` repos, which simplifies maintenance


.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+
+    - uses: hecrj/setup-rust-action@v1
+      with:
+        rust-version: ${{ matrix.rust_version }}
+
+    - name: Check documentation
+      run: cargo doc --no-deps --document-private-items
+
+    - name: Build
+      run: cargo build --verbose
```
kchibisov, line 47: We don't need `Build` just test.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,50 @@
...
+  CARGO_INCREMENTAL: 0
+  RUSTFLAGS: "-Cdebuginfo=0 --deny=warnings"
+  RUSTDOCFLAGS: "--deny=warnings"
+
+jobs:
+  fmt:
+    name: Check Formatting
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
```
MarijnS95, line 19: Nit: Arrays should be indented, also below.

kchibisov (changes requested):

PR #101, opened 2022-08-11
Specify MSRV

madsmtm:
Note that the intention here is that bumping MSRV is _not_ a breaking change! An example:
- We release `v0.5.1` with the `rust-version` field specified.
- Later on, we change something such that a newer Rust version is required; we bump `rust-version`, and release `v0.5.2`
- Users using Rust 1.56 will automatically choose `v0.5.1`

EDIT: Forget that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust_version: [stable, nightly]
+        rust_version: ["1.56", stable, nightly]
 
     steps:
       - uses: actions/checkout@v2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.5.1 (2022-08-11)
+
+* Add the `rust-version` field.
+
 ## 0.5.0 (2022-07-14)
 
 * **Breaking:** The `RawWindowHandle` variants were split into `RawDisplayHandle` and `RawWindowHandle`.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -2,13 +2,14 @@
 name = "raw-window-handle"
 version = "0.5.0"
 authors = ["Osspial <osspial@gmail.com>"]
-edition = "2018"
+edition = "2021"
 description = "Interoperability library for Rust Windowing applications."
 license = "MIT OR Apache-2.0 OR Zlib"
 repository = "https://github.com/rust-windowing/raw-window-handle"
 keywords = ["windowing"]
 readme = "README.md"
 documentation = "https://docs.rs/raw-window-handle"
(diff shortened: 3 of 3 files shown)


Lokathor:
That would be a great plan if rust-version affected version selection, but it does not.

kchibisov:
In general, I'm not against, but you'd need `1.57` version, given that `rust-version` was added in `1.57.0` IIRC.

madsmtm:
> That would be a great plan if rust-version affected version selection, but it does not.

Ah dammit, I thought I'd read that somewhere!

> In general, I'm not against, but you'd need `1.57` version, given that `rust-version` was added in `1.57.0` IIRC.

[It was stabilized in `1.56`](https://doc.rust-lang.org/nightly/cargo/reference/manifest.html#the-rust-version-field).

MarijnS95:
https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field @kchibisov It says 1.56 here.

As @Lokathor notes, this setting (afaik) only produces more helpful error messages, akin "your rust compiler version is too low" rather than bailing on unstabilized std interfaces or unknown/disallowed syntax.

Lokathor:
There's other MSRV elsewhere and I do not want to get into it much here but the short story is that a number of people are motivated to make `rust-version` affect resolution, so this might be fixed in the short or medium term.

**Until then** I'm happy to at least have the version set... but the rest of it about msrv bumps only being minor version bumps, that I'm not sure I'm willing to agree with.

Sound good to everyone?

MarijnS95:
@Lokathor When that happens, that'll be the MSRV on which resolution happens, meaning it'll take a bit longer for this to be relevant right?

Lokathor:
The user would need the newer `cargo`, but this 0.5.2 of the crate where we list a rust version would benefit without us having to further fix the 0.5 version series. So it's good to get this in and tracked now, even if it won't have an effect for a while.

(probably! the feature isn't precisely designed or implemented yet after all.)

Lokathor:
I also don't predict *much* need to move with msrv for this crate long term since the idea is to just be a data carrier in as plain a way as possible. Ideally once things are set we stop touching them, and new window managers come out rarely.

kchibisov:
The `rust-version` is not that of use, however what is helpful is testing msrv on CI, the `rust-version` is more of a hint now wrt what we're using.

kchibisov:
The only bump that will happen is when `c` types will be stable in rustc. So we can have zero deps crate.

madsmtm:
> The only bump that will happen is when `c` types will be stable in rustc. So we can have zero deps crate.

Which will be possible in 1.64: https://github.com/rust-lang/rust/issues/94501

Lokathor (approved):

Osoted56:
Me hackearon la cuenta verifiquen mi ip

Issue #92, opened 2022-07-03
Decouple display and window handles


kchibisov:
When user wants to deal with platform specific code they most of the time need raw display, like e.g. wl_display. However RawWindowHandle provides with both of them.

I'd suggest to logically split them into `RawDisplayHandle` and `RawWindowHandle`, where `RawDisplayHandle` would have only reference to display and `RawWindowHandle` could have only window, like `wl_surface` or be the way it is right now?

The issue I'm having is that I want Display on every platform, but I don't want surface, since the surface isn't yet created.

For example winit could implement `raw_display_handle` on event loop, so I can pass to e.g. glutin.

maroider:
From what I understand, this would make sense for `WaylandHandle.display`, `XcbHandle.connection`, and `XlibHandle.connection`? I'm not sure if there's something analogous on other platforms.

kchibisov:
For something like android/macOS it would make sense to have plain `Android` and `Macos`, etc. Since on those platforms you don't have those. Also with `Gbm(device)`.

So it'll work the same way it works right now, it just prevent me from creating `RawDisplay` right inside `glutin`.

maroider:
> So it'll work the same way it works right now, it just prevent me from creating `RawDisplay` right inside `glutin`.

Alright. I'm not familiar enough with `glutin` to understand why this would be useful, but I'm assuming this has something to do with decoupling `glutin` from `winit`?

kchibisov:
>Alright. I'm not familiar enough with glutin to understand why this would be useful, but I'm assuming this has something to do with decoupling glutin from winit?

To create EGLDisplay or GLXDisplay you need raw display handle, nothing more. This display allows you to create objects, such as EGLContext, EGLConfig, EGLSurface, etc. The RawWindowHandle is only required for EGLSurface, and not all the surfaces actually, since e.g. PBuffers don't need it at all.

I'd assume the same applies to vulkan or e.g. softbuffer(not the current impl of softbuffer in particular).

MarijnS95:
> I'd assume the same applies to vulkan

I don't recall there being any concept of a "Display" connection (needed) to do this. As long as you can enumerate a Vulkan device one can create/allocate a texture for "rendertarget" purposes.

You'll need both to create a surface _on a window_ though:

https://github.com/ash-rs/ash/blob/master/ash-window/src/lib.rs

[...]

PR #106, opened 2023-01-03
Expand documentation of RawDisplayHandle

atsheehan:
I came to this crate after reading about the recent changes to [`glutin`][0]. At a glance I didn't understand the difference between `RawWindowHandle` and `RawDisplayHandle` (or even what a display was), and the enums looked mostly the same.

After reading through this [issue][1] and digging through the structs, I have a better understanding of what each type represents, but I thought it would be useful to add a line clarifying the distinction between the two. I also added some examples of what external types may be represented for a few platforms.

[0]: https://github.com/rust-windowing/glutin/releases/tag/v0.30.0
[1]: https://github.com/rust-windowing/raw-window-handle/issues/92

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -18,6 +18,12 @@
 //! added without breaking backwards compatibility. Each struct provides an `empty` method that may
 //! be used along with the struct update syntax to construct it. See each specific struct for
 //! examples.
+//!
+//! ## Display Handles
+//!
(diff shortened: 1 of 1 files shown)


Lokathor (approved):

Lokathor:
I love more docs, but rather than merge this immediately I'll leave it open for others to have a chance to speak up about the particular wording.

madsmtm (approved):

PR #107, opened 2023-02-16
Remove `cty` dependency

### LINKED ISSUE #103: Remove `cty` dep
Given that rust 1.64 got released we can get rid of it.

kchibisov:
This bumps MSRV to 1.64 along the way.

Fixes #103.

--

It seems that `wgpu`(already 1.64) and friends are moving towards the `1.64`. In general rust considers MSRV bump as not breaking for 'stable - 2' (including stable).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,7 +31,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        rust_version: ["1.56", stable, nightly]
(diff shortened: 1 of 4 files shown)


Lokathor (approved):

Lokathor:
I sure don't know about that "stable -2" part, see the libc debate, but yeah this is fine.

I expect this won't get released until the other PR that makes the trait fallible is also done, so this'll be part of a breaking release anyway.

PR #109, opened 2023-02-18
Implement `From` for `Raw{Window,Display}Handle`

### LINKED ISSUE #108: Implement `From` for `Window`/`Display` handles
The use case is to save extra imports, like

```rust
fn raw_display_handle() -> RawDisplayHandle {
    let mut handle = WaylandDisplayHandle::empty();
    handle.wayland_display = display;
    handle.into()
}
```

kchibisov:
Fixes #108.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 
+* Implemented `From` for `RawWindowHandle` and `RawDisplayHandle`
 * The `rust-version` was bumped to `1.64`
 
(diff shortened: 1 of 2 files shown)


kchibisov:
I wonder whether we can test it somehow when new enum variant is added, I don't think we can?

Lokathor:
I don't think that there's any way to test that this covers all enum variants, but if a From impl is ever missing i'm sure someone will tell us and we can add the impl easily enough.

Lokathor (approved):

PR #110, opened 2023-03-17
feat: Add safer borrowed handles

notgull:
Pretty much a copy-paste of my `window-handle` repo to this crate. See [here](https://github.com/rust-windowing/winit/pull/2744#issuecomment-1474304925) for rationale.

By the way, would it be possible to add me to the @rust-windowing/raw-window-handle team so that I can maintain this code?

### DIFF
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -1,0 +1,278 @@
+//! Borrowable window handles based on the ones in this crate.
+//!
+//! These should be 100% safe to pass around and use, no possibility of dangling or
+//! invalidity.
+
+use core::fmt;
+use core::marker::PhantomData;
+
+use crate::{
+    HasRawDisplayHandle, HasRawWindowHandle, RawDisplayHandle, RawWindowHandle,
+};
+
+/// The application is currently active.
+///
+/// This structure is a token that indicates that the application is
+/// not presently suspended. It is used to ensure that the window handle
+/// is only used when the application is active.
+///
+/// It is safe to create this token on platforms where the application
+/// is guaranteed to be active, such as on desktop platforms. On Android
+/// platforms, the application may be suspended, so this token must be
+/// either created with `unsafe` or derived from a `HasDisplayHandle`
+/// type.
+pub struct Active<'a> {
+    _marker: PhantomData<&'a *const ()>,
+}
+
+impl fmt::Debug for Active<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.write_str("Active { .. }")
+    }
+}
+
+impl<'a> Active<'a> {
+    /// Create a new active token.
+    ///
+    /// # Safety
+    ///
+    /// The application must not be `Suspend`ed.
+    ///
+    /// # Examples
+    ///
+    /// ```
+    /// use raw_window_handle::Active;
+    ///
+    /// // SAFETY: The application is active.
+    /// let active = unsafe { Active::new_unchecked() };
+    /// ```
+    pub unsafe fn new_unchecked() -> Self {
+        Self {
+            _marker: PhantomData,
+        }
+    }
+
+    /// Create a new active token on a system where the application is
+    /// guaranteed to be active.
+    ///
+    /// On most platforms, there is no event where the application is
+    /// suspended, so there are no cases where this function is unsafe.
+    ///
+    /// ```
+    /// use raw_window_handle::Active;
+    ///
+    /// let with_active = |active: Active<'_>| {
+    ///     /* ... */
+    /// };
+    ///
+    /// // Only use this code-path on non-android platforms.
+    /// #[cfg(not(target_os = "android"))]
+    /// {
+    ///     let active = Active::new();
+    ///     with_active(active);
+    /// }
+    ///
+    /// // Only use this code-path on android platforms.
+    /// #[cfg(target_os = "android")]
+    /// {
+    ///     if application_is_active() {
+    ///         let active = unsafe { Active::new_unchecked() };
+    ///         with_active(active);
+    ///     }
+    /// }
+    /// # fn application_is_active() -> bool { false }
+    /// ```       
+    #[cfg(not(target_os = "android"))]
+    #[cfg_attr(docsrs, doc(cfg(not(target_os = "android"))))]
+    pub fn new() -> Self {
+        // SAFETY: The application is guaranteed to be active.
+        unsafe { Self::new_unchecked() }
+    }
+}
+
+/// A display that acts as a wrapper around a display handle.
+///
+/// # Safety
+///
+/// The safety requirements of [`HasRawDisplayHandle`] apply here as
+/// well. To clarify, the [`DisplayHandle`] must contain a valid window
+/// handle for its lifetime. In addition, the handle must be consistent
+/// between multiple calls barring platform-specific events.
+///
+/// In addition, the `active` function must only return an `Active`
+/// token if the application is active.
+///
+/// Note that these requirements are not enforced on `HasDisplayHandle`,
+/// rather, they are enforced on the constructors of [`Active`] and
+/// [`DisplayHandle`]. This is because the `HasDisplayHandle` trait is
+/// safe to implement.
+///
+/// [`HasRawDisplayHandle`]: crate::HasRawDisplayHandle
+pub trait HasDisplayHandle {
+    /// Get a token indicating whether the application is active.
+    fn active(&self) -> Option<Active<'_>>;
+
+    /// Get a handle to the display controller of the windowing system.
+    fn display_handle<'this, 'active>(
+        &'this self,
+        active: &'active Active<'_>,
+    ) -> DisplayHandle<'this>
+    where
+        'active: 'this;
+}
+
+/// The handle to the display controller of the windowing system.
+///
+/// Get the underlying raw display handle with the `HasRawDisplayHandle` trait.
+#[repr(transparent)]
+#[derive(PartialEq, Eq, Hash)]
+pub struct DisplayHandle<'a> {
+    raw: RawDisplayHandle,
+    _marker: PhantomData<&'a *const ()>,
+}
+
+impl fmt::Debug for DisplayHandle<'_> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_tuple("DisplayHandle").field(&self.raw).finish()
+    }
+}
+
+impl<'a> Copy for DisplayHandle<'a> {}
+impl<'a> Clone for DisplayHandle<'a> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<'a> DisplayHandle<'a> {
+    /// Borrow a `DisplayHandle` from a `RawDisplayHandle`.
+    ///
+    /// # Safety
+    ///
+    /// The `RawDisplayHandle` must be valid for the lifetime and the
+    /// application must be `Active`. See the requirements on the
+    /// [`HasDisplayHandle`] trait for more information.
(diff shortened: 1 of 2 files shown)


Lokathor:
I'd be happy to have you on the team but I don't think I personally have permissions to edit the team. One of the higher level rust-windowing admins should be able to add you.

kchibisov:
>By the way, would it be possible to add me to the @rust-windowing/raw-window-handle team so that I can maintain this code?

Done.

notgull:
Thanks @kchibisov! I'll merge this patch in.

Lokathor:
@notgull can you also do a follow up PR to put this in the changelog so i can just pull the latest and release the patch once I'm free later?

rib:
Hey, sorry for being late to the party here. I'm just trying to follow the recent changes here and get a clearer understanding.

Here are a few initial comments I made on related issue #111 where I was pinged https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1493301738

I'm not sure that there was a safety issue on Android here - assuming a valid reference to an `ANativeWindow` is held.

Android apps need to be aware to recreate render surfaces when they resume but it was possible to handle that safely with the previous raw-window-handle + Winit API (E.g. here are some examples that work on Android and are also portable to desktop window systems: https://github.com/rust-mobile/rust-android-examples)

I'm maybe missing a full understanding currently of how consumers of this are intended to handle the introduced `Active` state here.

Are consumers now required to poll when a display handle becomes active/inactive and maybe automatically re-create surfaces when transitioning to active, and NOP when inactive?

My gut feeling at the moment is that Android applications should be driven by the app lifecycle events directly and instead of having handle consumers poll for lifecycle changes and automatically re-creating surfaces it seems like we should give out new handles to consumers when resumed/re-activated.

There can surely be lots of, potentially app-specific, details involved in creating graphics API surfaces from a handle so I'm very unclear if the implication with this design is that consumers polling for `active` state changes need to always be able to automatically/transparently re-create a surface that was equivalent to a previously created/configured surface? I don't think I'd expect that to always be feasible.

notgull:
> I'm maybe missing a full understanding currently of how consumers of this are intended to handle the introduced `Active` state here.
> 
> Are consumers now required to poll when a display handle becomes active/inactive and maybe automatically re-create surfaces when transitioning to active, and NOP when inactive?
> 
> My gut feeling at the moment is that Android applications should be driven by the app lifecycle events directly and instead of having handle consumers poll for lifecycle changes and automatically re-creating surfaces it seems like we should give out new handles to consumers when resumed/re-activated.

The theory is that consumers would actually hold references to the underlying `HasWindowHandle` object (such as the `winit::window::Window`). Under the hood, the lifecycle system would call `Active::set_inactive()` and `Active::set_active()` in order to set whether or not it's valid to create new window handles.

There isn't really any kind of "polling" here; when you want to access the window you check whether or not the application is suspended. While you hold the window handle, the application won't enter the suspended state until you stop holding it.

> I'm not sure that there was a safety issue on Android here - assuming a valid reference to an `ANativeWindow` is held.

I'm not sure that the reference stays valid forever, as the [docs](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1a6aaafbbfae4c0ac066b9d61ebb3ab97f) say:

> You MUST ensure that you do not touch the window object after returning from this function

The docs also seem to imply that the pointer could change in between the `onNtiveWindowDestroyed` and `onNativeWindowCreated`.

MarijnS95:
> There isn't really any kind of "polling" here; when you want to access the window you check whether or not the application is suspended. While you hold the window handle, the application won't enter the suspended state until you stop holding it.

This sounds problematic. As mentioned in https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1493408521 the typical graphics API doesn't "want to access the window" to check whether it may be suspended at random: this (creating a `Surface`/`Swapchain`) is a one-off setup at the start of an app, and the lifecycle events currently provided by `ndk-glue`/`android-activity` and `winit` inform the user that it's time to clean up their _implicit_ references to `ANativeWindow` by destroying their `Surface`/`Swapchain`.

The only odd case for add-hoc getting the window/handle is to read the size, or [call `ANativeWindow_lock()` for CPU-based pixel blitting](https://github.com/rust-windowing/softbuffer/issues/44) (the latter will happen on the `NativeWindow` struct directly though, not via its abstracted window handle).

rib:
> > I'm maybe missing a full understanding currently of how consumers of this are intended to handle the introduced `Active` state here.
> > Are consumers now required to poll when a display handle becomes active/inactive and maybe automatically re-create surfaces when transitioning to active, and NOP when inactive?
> > My gut feeling at the moment is that Android applications should be driven by the app lifecycle events directly and instead of having handle consumers poll for lifecycle changes and automatically re-creating surfaces it seems like we should give out new handles to consumers when resumed/re-activated.
> 
> The theory is that consumers would actually hold references to the underlying `HasWindowHandle` object (such as the `winit::window::Window`). Under the hood, the lifecycle system would call `Active::set_inactive()` and `Active::set_active()` in order to set whether or not it's valid to create new window handles.
> 
> There isn't really any kind of "polling" here; when you want to access the window you check whether or not the application is suspended. While you hold the window handle, the application won't enter the suspended state until you stop holding it.

I'm not sure how this will be manageable. If you create a surface, such as an EGLSurface that's effectively holding on to the window handle indefinitely. Creating a surface is going to hold a reference to the `ANativeWindow`, so conceptually I think this lock needs to be held all the time that the surface exists - not just temporarily in some event handler callbacks. If that were the case though you'd never be able to acquire the lock to suspend?

E.g. see docs for creating an Vulkan surface which say how it takes a reference to the `ANativeWindow`: https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/vkCreateAndroidSurfaceKHR.html#_c_specification

> 
> > I'm not sure that there was a safety issue on Android here - assuming a valid reference to an `ANativeWindow` is held.
> 
> I'm not sure that the reference stays valid forever, as the [docs](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1a6aaafbbfae4c0ac066b9d61ebb3ab97f) say:
> 
> > You MUST ensure that you do not touch the window object after returning from this function
> 
> The docs also seem to imply that the pointer could change in between the `onNtiveWindowDestroyed` and `onNativeWindowCreated`.

The `ANativeWindow` is effectively the backend for `android.view.Surface` and you can also explicitly get an `ANativeView` from a JNI reference to a surface.

You could e.g. call `ANativeWindow_fromSurface` and get an `ANativeWindow` outside of the `onNativeWindowCreated/Destroyed` protocol - in which case that documentation could be said to not apply. The problem is that that documentation isn't really for `ANativeWindow` itself - it's for `ANativeActivity`.

Your app won't work properly if you do just continue trying to use the window after the `NativeActivity` has informed you that its "terminated" so it's at least prudent that applications shouldn't do that but I'm not sure this is the same as saying it's going to be unsafe/unsound to access the `ANativeWindow` API while you own a reference to it.

Considering how the Surface Java API is a thin layer over `ANativeWindow`: https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/view/Surface.java i'd be surprised if `ANativeWindow` wasn't always safe to access while you have a valid reference.

As another point of reference it's also possible to acquire an `ANativeWindow` without using `NativeActivity`. E.g. in the implementation of the `GameActivity` backend of `android-activity` you can see how the backend simply creates an `ANativeWindow` from a surface and which it's finished with it then it `_release`s it via its ref count: https://github.com/rust-mobile/android-activity/blob/36ddfaa9cea7af9377aa43e0db6cc56e80ccc234/android-activity/game-activity-csrc/game-activity/GameActivity.cpp#L335

At least anecdotally I can also say that when I've accessed `ANativeWindow` I've not seen that lead to an abort/segfault etc - it's just resulted in a black screen from not being able to render to the surface any more.

notgull:
> As mentioned in [#111 (comment)](https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1493408521) the typical graphics API doesn't "want to access the window" to check whether it may be suspended at random: this (creating a `Surface`/`Swapchain`) is a one-off setup at the start of an app, and the lifecycle events currently provided by `ndk-glue`/`android-activity` and `winit`

The idea is that, before you call any `unsafe` code that utilizes the event handle, you would run:

```rust
let _guard = match self.window_type.window_handle() {
    Err(HandleError::Inactive) => return Err(MyError::WindowWasInactive),
    Ok(handle) => {
        if handle != self.raw_window_handle_thats_registered_in_gl_or_whatever {
            return Err(MyError::UserDidntDropUsWhenWeRequestedIt);
        }

        handle
    }
};
```

The idea is that:

- The `_guard` makes sure the handle isn't dropped out from under us.
- If the handle changed, we should invalidate ourselves, so just error out.
- If we're called while the window is inactive, error out.

As far as I know, this is the best way of delivering the notification that the window is suspended in a way that can be done 100% safely by the end user.

(Now that I think about it, the above code might be useful to add to `HasWindowHandle` as an extension trait...)

> The `ANativeWindow` is effectively the backend for `android.view.Surface` and you can also explicitly get an `ANativeView` from a JNI reference to a surface.
> 
> You could e.g. call `ANativeWindow_fromSurface` and get an `ANativeWindow` outside of the `onNativeWindowCreated/Destroyed` protocol - in which case that documentation could be said to not apply. The problem is that that documentation isn't really for `ANativeWindow` itself - it's for `ANativeActivity`.

This is an interesting point that I wasn't aware of. In this case, I think the current API can accommodate this, though; we just add documentation to the [`ActiveHandle::new_unchecked`](https://github.com/rust-windowing/raw-window-handle/blob/4abe033f26df8ce00ab49f983e525f982ed3f49e/src/borrowed.rs#L172-L194) function that it's allowed if the `ANativeWindow` was derived from an external JNI reference.

> I'm not sure how this will be manageable. If you create a surface, such as an EGLSurface that's effectively holding on to the window handle indefinitely. Creating a surface is going to hold a reference to the `ANativeWindow`, so conceptually I think this lock needs to be held all the time that the surface exists - not just temporarily in some event handler callbacks. If that were the case though you'd never be able to acquire the lock to suspend?

I think that the system I mentioned above should counteract this.

rib:
> > I'm not sure how this will be manageable. If you create a surface, such as an EGLSurface that's effectively holding on to the window handle indefinitely. Creating a surface is going to hold a reference to the `ANativeWindow`, so conceptually I think this lock needs to be held all the time that the surface exists - not just temporarily in some event handler callbacks. If that were the case though you'd never be able to acquire the lock to suspend?
> 
> I think that the system I mentioned above should counteract this.

I'm not sure that it does - or I didn't follow how it does.

this protocol is concerned with (not) letting the app touch the `ANativeWindow` outside of `onNativeWindowCreated/Destroyed` - due to a concern that it may be unsafe.

If an `EGLSurface/VkSurface` takes a reference to an `ANativeWindow` then that effectively continues to hold the handle (not from a Rust pov necessarily but it holds a reference to the thing we're trying to control) outside of the guard. You don't necessarily know what actions might lead a graphics API to poke at a surface.

I don't see how the transient guard will be able to affect the retention of surface references passed on to C APIs that are going to be held for long periods of time.

What's missing is that Android apps need to go and explicitly drop all graphics surfaces that could be holding on to an `ANativeWindow`, but I think that's a higher-level concern that probably can't be done automatically at the `raw-window-handle` level.

MarijnS95:
> Your app won't work properly if you do just continue trying to use the window after the `NativeActivity` has informed you that its "terminated"

> What's missing is that Android apps need to go and explicitly drop all graphics surfaces that could be holding on to an `ANativeWindow`

For completeness we deduced that holding on to `ANativeWindow` - with the appropriate refcount increase - is totally fine, but rendering to it won't have any effect. It's not just about dropping the graphics surface on "suspend", but also about _picking up a new, fresh `ANativeWindow` on "resume"_ :slightly_smiling_face:

PR #112, opened 2023-03-17
Add #110 to the changelog

notgull:
https://github.com/rust-windowing/raw-window-handle/pull/110#issuecomment-1474425418

Should I also bump the version?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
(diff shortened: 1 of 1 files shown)


Lokathor:
I personally use cargo-release to do releases, which can bump the version as part of the release (eg: `cargo release patch --execute`), so that's not necessary.

I've never been sure if people expect the main branch cargo file to have the version bump already checked in during that version's development, or only bump at the moment of actual release. Unless there's some general "rust projects should do this" guidance in the rust-lang organization somewhere then bumping at the last moment seems okay with me.

Lokathor (approved):

PR #115, opened 2023-03-20
feat: Implement HasDisplayHandle and HasWindowHandle on reference types

notgull:
Forgot to do this the first time around. Makes sure that these traits can be used from reference as well.

### DIFF
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -119,6 +119,76 @@ pub trait HasDisplayHandle {
         'active: 'this;
 }
 
+impl<T: HasDisplayHandle + ?Sized> HasDisplayHandle for &T {
+    fn active(&self) -> Option<Active<'_>> {
+        (**self).active()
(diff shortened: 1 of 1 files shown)


src/borrowed.rs
```diff
@@ -119,6 +119,76 @@ pub trait HasDisplayHandle {
...
 
+impl<T: HasDisplayHandle + ?Sized> HasDisplayHandle for &T {
+    fn active(&self) -> Option<Active<'_>> {
+        (**self).active()
+    }
+
+    fn display_handle<'this, 'active>(
+        &'this self,
+        active: &'active Active<'_>,
+    ) -> DisplayHandle<'this>
```
ogoffart, line 130: Based on the discussion in https://github.com/rust-windowing/raw-window-handle/pull/104 , shouldn't this return a `Option<DisplayHandle<'this>>` ?

PR #114, opened 2023-03-20
docs: Improve documentation for safe window handles

notgull:
Improves documentation for the new types and traits in the `borrowed.rs` file.

I did some research into Android lifecycles, so I'm pretty sure now that the `Active` trick works and why it does. I also add notes on window ID invalidation to the window handle part.

cc #111

Issue #111, opened 2023-03-17
Refine the safe window handle API


notgull:
In #110 I added safer versions of the window handles. However, we might need to make sure they work with all platforms, present and future, before we release them.

Platform specific issues that I'm aware of:

- The Android application can be randomly invalidated, making it undefined behavior to use the handles. However, we use `Active` to work around that by only allowing usage of the handles while the application is active.
- Windows in X11 (and Wayland, I think?) can be deleted at any point using [100% safe code](https://docs.rs/x11rb/latest/x11rb/rust_connection/struct.RustConnection.html). I don't think this is an issue, since I don't think any undefined behavior can result from this; worst comes to worse you get a `BadWindow` error. However, I'm not sure if there are any implications on the GPU end that we should be aware of.

kchibisov:
>and Wayland, I think

Not really, it'll raise a protocol error if you do that before the role object and it is not exposed. So while you can issue a drop, you'll kill the entire app.

ids1024:
Well, on X you could just delete a window in safe code, in a different process, on a different machine connected to the same X server, right? You can't really rely on anything behaving particularly consistently on X unless all clients are cooperative, but it should still be safe at least.

On Wayland, yeah, if you destroy the role object and other things the protocol states should be destroyed first. Then I think you could get send requests with an object ID that is no longer valid? Which probably means a protocol error, unless it has been re-used for another surface.

The documentation should probably specifically call out these things the API cannot guarantee.

ids1024:
The `Active` part of the API is a bit ugly and awkward, but I'm not familiar enough with this aspect of Android to know precisely what this is necessary for (without the restriction, would this violate soundness somehow, or just cause other issues?), whether this correctly addresses it, or if there's a better solution...

Lokathor:
On Windows, someone can just plain delete your window on you. You don't need to have even shared the HWND with any other code, there's functions to enumerate all visible windows for example. From a separate process eve. (If you use a window handle to a deleted window it's supposed to just give "invalid handle" errors, but it's a hard to test edge case so i wouldn't be surprised if possible bugs occurred.)

I'm not sure how this "active" thing interacts with that

ids1024:
Ah, that's a thing on Windows too? Even from another process? I guess it makes sense since the core of the win32 API for dealing with windows is largely the same as the win16 API included in Windows 1.01 in 1985, so it's not particularly more modern than X...

[...]

PR #116, opened 2023-03-24
Change the "Active" struct to be sound

notgull:
The previous implementation is unsound in several cases. This PR ensures that window handles don't outlive the application.

Draft until I can verify that it actually works. Supersedes #114

### DIFF
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -335,13 +335,13 @@ impl<'a> HasDisplayHandle for DisplayHandle<'a> {
 /// the handle.
 ///
 /// Note that this guarantee only applies to *pointers*, and not any window ID types in the handle.
-/// This includes Window IDs (XIDs) from X11, `HWND`s from Win32, and the window ID for web platforms.
-/// There is no way for Rust to enforce any kind of invariant on these types, since:
+/// This includes Window IDs (XIDs) from X11 and the window ID for web platforms. There is no way for
+/// Rust to enforce any kind of invariant on these types, since:
 ///
 /// - For all three listed platforms, it is possible for safe code in the same process to delete
 ///   the window.
-/// - For X11 and Win32, it is possible for code in a different process to delete the window.
-/// - For X11, it is possible for code on a different *machine* to delete the window.
+/// - For X11, it is possible for code in a different process to delete the window. In fact, it is 
+///   possible for code on a different *machine* to delete the window.
 ///
 /// It is *also* possible for the window to be replaced with another, valid-but-different window. User
 /// code should be aware of this possibility, and should be ready to soundly handle the possible error
@@ -388,33 +388,12 @@ impl<H: HasWindowHandle + ?Sized> HasWindowHandle for alloc::sync::Arc<H> {
     }
 }
 
-/// The error type returned when a handle cannot be obtained.
-#[derive(Debug)]
-#[non_exhaustive]
-pub enum HandleError {
-    /// The handle is not currently active.
-    ///
-    /// See documentation on [`Active`] for more information.
-    Inactive,
-}
-
-impl fmt::Display for HandleError {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
-            Self::Inactive => write!(f, "the handle is not currently active"),
-        }
-    }
-}
-
-#[cfg(feature = "std")]
-impl std::error::Error for HandleError {}
-
 /// The handle to a window.
 ///
 /// This is the primary return type of the [`HasWindowHandle`] trait. All *pointers* within this type
 /// are guaranteed to be valid and not dangling for the lifetime of the handle. This excludes window IDs
-/// like XIDs, `HWND`s, and the window ID for web platforms. See the documentation on the
-/// [`HasWindowHandle`] trait for more information about these safety requirements.
+/// like XIDs and the window ID for web platforms. See the documentation on the [`HasWindowHandle`] 
+/// trait for more information about these safety requirements.
 ///
 /// This handle is guaranteed to be safe and valid. Get the underlying raw window handle with the
 /// [`HasRawWindowHandle`] trait.
@@ -473,6 +452,27 @@ impl HasWindowHandle for WindowHandle<'_> {
         Ok(self.clone())
     }
 }
+
+/// The error type returned when a handle cannot be obtained.
+#[derive(Debug)]
+#[non_exhaustive]
+pub enum HandleError {
+    /// The handle is not currently active.
+    ///
+    /// See documentation on [`Active`] for more information.
+    Inactive,
+}
+
+impl fmt::Display for HandleError {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::Inactive => write!(f, "the handle is not currently active"),
+        }
+    }
+}
+
+#[cfg(feature = "std")]
+impl std::error::Error for HandleError {}
 
 /// ```compile_fail
 /// use raw_window_handle::{Active, DisplayHandle, WindowHandle};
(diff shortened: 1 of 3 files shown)


src/borrowed.rs
```diff
@@ -1,197 +1,129 @@
...
-    /// // SAFETY: The application is active.
-    /// let active = unsafe { Active::new_unchecked() };
-    /// ```
-    pub unsafe fn new_unchecked() -> Self {
-        Self {
-            _marker: PhantomData,
-        }
+    /// The application must actually be inactive.
+    pub unsafe fn set_inactive(&self) {
+        self.0.set_inactive()
```
Imberflur, line 52: Does `set_inactive` need to be `unsafe`? Afaict, for soundness it is only necessary that this is not falsely marked as active.

notgull: Hmm, you might be right. However, I think it might work in either direction; could we make `set_active()` safe and `set_inactive()` unsafe, since `set_inactive()` would only really be called on Android?

notgull: Yeah, you're right. I'll patch it.

notgull:
android-glutin-triangle

Looks like the `glutin` example works on Android, at least from a glance. I think we're ready to start merging and releasing this.

Lokathor:
Just having a look on the bus, things seem pretty good.

I'd like for there to be more documentation, perhaps at the module level, about how a person writing a system lib should approach things (eg: winit/sdl2 folks), and how a person writing a graphics lib (eg: wgpu/ash folks) should approach things. I think the ideal is that people *other than* those two groups shouldn't have to look too carefully in here? Just pass a winit thing to a wgpu thing? So the system lib audience and graphics lib audience seem like our main users.

I wouldn't say that this is a blocking concern, but "good docs" is the second most important part of a crate.

Lokathor (approved):
I'm gonna approve this now, and leave it up to you notgull on if you want to add more usage docs now or not. The alternative is we just merge it and we can always add more docs as a patch if people end up having specific questions.

notgull:
> I'd like for there to be more documentation, perhaps at the module level, about how a person writing a system lib should approach things (eg: winit/sdl2 folks), and how a person writing a graphics lib (eg: wgpu/ash folks) should approach things. I think the ideal is that people _other than_ those two groups shouldn't have to look too carefully in here? Just pass a winit thing to a wgpu thing? So the system lib audience and graphics lib audience seem like our main users.

Sounds good! I've added some more documentation, with some specifically targeting `winit`/`wgpu` developers.

Lokathor:
I think for the doc links to work out approximately correctly you'll have to put in links to `https://docs.rs/crate_name` (or make them not links)

src/borrowed.rs
```diff
@@ -61,19 +171,47 @@ impl Active {
...
 }
 
 /// A display that acts as a wrapper around a display handle.
 ///
+/// Objects that implement this trait should be able to return a [`DisplayHandle`] for the display
+/// that they are associated with. This handle should last for the lifetime of the object, and should
+/// return an error if the application is inactive.
+///
+/// Implementors of this trait will be windowing systems, like [`winit`] and [`sdl2`]. These windowing
+/// systems should implement this tait on types that already implement [`HasRawDisplayHandle`]. It
```
Imberflur, line 204: typo: tait -> trait

notgull: Thanks for the catch!

It might be worthwhile to run a spellchecker on the CI for this crate.

src/borrowed.rs
```diff
@@ -236,108 +305,123 @@ unsafe impl HasRawDisplayHandle for DisplayHandle<'_> {
...
+/// All pointers within the resulting [`WindowHandle`] must be valid and not dangling for the lifetime of
+/// the handle.
+///
+/// Note that this guarantee only applies to *pointers*, and not any window ID types in the handle.
+/// This includes Window IDs (XIDs) from X11, `HWND`s from Win32, and the window ID for web platforms.
+/// There is no way for Rust to enforce any kind of invariant on these types, since:
+///
+/// - For all three listed platforms, it is possible for safe code in the same process to delete
+///   the window.
+/// - For X11 and Win32, it is possible for code in a different process to delete the window.
```
Imberflur, line 343: Given https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1484369947 is this accurate for Win32?

Lokathor: The proper way to do it is for the other process to send `WM_CLOSE` to your window (a "close request" event), and then your window can respond to that on its own thread, and usually yes the response is to call `DestoryWindow`, but it might be to open a message box (eg: "are you you want to quit without saving?"), or something else such as posting a quit event to the thread's event queue.

notgull: > The proper way to do it is for the other process to send `WM_CLOSE` to your window (a "close request" event)

Yes, this is exposed in `winit` through the `CloseRequested` event. Anyways, it's probably a good idea to remove this section anyways, since `DestroyWindow` can't be called arbitrarily.

rib:
heya, just continuing to follow the trail of recent changes here...

Just because I see that Glutin was used to smoke test this change above - does that mean that you were seeing a problem with Glutin on Android before?

As far as I know, this Glutin example had been working for me on Android: https://github.com/rust-mobile/rust-android-examples/tree/main/na-winit-glutin (working across suspend / resume and cross-platform too)

The main consideration that was required for Android apps was that they need to re-create render surfaces when they resume. One portable way of handling this is that apps can drop their surface state when they Suspend (only ever happens on Android / iOS) and they can lazily create the render surfaces when they Resume (in Winit, all platforms get a Resume event).

I have to admit I'm currently a bit daunted by the new `Active` state tracking here and wondering about the implications of the `RwLock` that's involved in the `Active` tracking for Android. (without scrutinizing the details, I'm just initially worried about whether this could dead lock an event loop trying to suspend or resume)

I've had an initial look at the draft changes to use this in Glutin but I don't know that I entirely followed how the active status is handled (I just noticed that when `window.window_handle()` was being queried from Winit, then the code was expecting that it would be active I think:

```rust
let raw_window_handle =
        window.as_ref().map(|window| window.window_handle()).transpose().expect("Inactive");
```

and I saw that various places (like `create_window_surface`) would return an `Err` `Result` I think if a handle wasn't active but does that mean that this change is purely about avoiding accessing a handle while inactive? (i.e. it doesn't introduce special handling of the inactive case to try and hide that state from the app)

My current understanding is that if you want to write an application for Android based on this change then it's still your responsibility to only try and create a surface while 'active' and this doesn't try to make APIs such as Glutin automatically poll the active status and lazily [re]create surfaces (that's what I originally guessed this change might be aiming for).

If that's the case then I'm not currently sure if this change was required if there wasn't really a safety issue with accessing inactive `ANativeWindow`s on Android?

As far as I know an `ANativeWindow` should remain safe to access so long as you hold a reference - it's just that the `BufferQueue` behind it isn't going to hand out new buffers for rendering once it's been "terminated".

Again, sorry for being late with asking questions here, since this has already been merged, but it feels like quite a significant change which I think is likely to affect my usage and I'm hoping to get a clearer understanding and also double check what's required here.

notgull:
> Just because I see that Glutin was used to smoke test this change above - does that mean that you were seeing a problem with Glutin on Android before?

No, it was mostly just to verify that nothing serious had been broken by my changes. I would've used `softbuffer` but it doesn't support Android yet, see rust-windowing/softbuffer#44

> I have to admit I'm currently a bit daunted by the new `Active` state tracking here and wondering about the implications of the `RwLock` that's involved in the `Active` tracking for Android. (without scrutinizing the details, I'm just initially worried about whether this could dead lock an event loop trying to suspend or resume)

I agree, this is definitely the "monster in the closet" for the current implementation; it introduces a potential deadlock that wasn't there before (i.e. someone is holding a window handle while waiting for events, while the event loop is blocked waiting for that user to drop a window handle).

I think that the [docs](https://developer.android.com/ndk/reference/struct/a-native-activity-callbacks#struct_a_native_activity_callbacks_1a6aaafbbfae4c0ac066b9d61ebb3ab97f) explicitly call out the fact that you shouldn't use a native window handle after `onNativeWindowDestroyed`:

> You MUST ensure that you do not touch the window object after returning from this function

MarijnS95:
@notgull let me know if you need a `softbuffer` implementation for Android, then I'll prioritize cleaning and merging the promised changes.

PR #117, opened 2023-03-30
Restrict compilation of borrowed handles on Android

notgull:
Guards borrowed handles behind either `std` or not being on Android. Should prevent this from being a breaking change.

Another possible way around this would be to add a `borrowed` feature.


Lokathor:
it's smoothest if we just turn this on for everyone without the cargo feature, since we'd only remove the cargo feature later. This way, the only people affected are android builders with no_std (who i suspect are very rare).

Lokathor (approved):

PR #118, opened 2023-04-04
Remove Android-specific platform differences

notgull:
Makes `Active` a no-op on all platforms.

cc @rib and @MarijnS95 Is this valid? I want to make extra sure, I don't want to screw this up twice


MarijnS95:
@notgull thanks for sending this in, I wouldn't call the original a screw-up outright but haven't caught up to the discussion in existing issues PRs (which should definitely be linked and explained as part of the context for _why_ this PR is making this change :wink:). Will take a closer look tomorrow, and we can always push users to not use the no-op via `#[deprecated]` without breaking changes. On that note I'm sorry for repeating the below if it has already been discussed:

It quite surprises me that these changes have been made and [pressured into a release](https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1490737912) within a timespan of about two weeks, without any apparent community discussion/review not to mention a comment from active Android+Rust platform maintainers (turns out @rib was on holiday, and I had a massive backlog to process after also being out of country for a conference). I get it, it's sometimes hard to not have a release while porting other crates in the ecosystem to the new thing, but using temporary `git` references (for example) allows the new API to settle for a while to get more eyes on it and work out the final kinks (for a critical crate like RWH). We used to do that for the `ndk` in `winit`, so that it could use the latest features and force the `ndk` to finally make a release with `winit` in lockstep. It does help that these are (or were...) all under the same GitHub organization back then, and now with RWH.
But that's just my 2cts on the process. Still, thanks again for working on this crate and coming up with a safe abstraction!

(And for the commit title: `s/no-ip/no-op`)

notgull:
> Will take a closer look tomorrow, and we can always push users to not use the no-op via `#[deprecated]` without breaking changes.

Would probably be for the best (although I doubt there are any extant users aside from my PRs).

> It quite surprises me that these changes have been made and [pressured into a release](https://github.com/rust-windowing/raw-window-handle/issues/111#issuecomment-1490737912) within a timespan of about two weeks, without any apparent community discussion/review not to mention a comment from active Android+Rust platform maintainers (turns out @rib was on holiday, and I had a massive backlog to process after also being out of country for a conference).

Yes, this was undeniably a mistake on my part with no excuses. At this stage, I'd been able to implement `Has[Display/Window]Handle` on most types in the ecosystem without any real friction. In addition, I'd posted about my initiative to [Reddit](https://www.reddit.com/r/rust/comments/121j6xu/a_proposal_for_safe_window_handles/) and had not received any real opposition to my design. My rationale was that at least one person familiar with Android would read my post and point out flaws if there were any. I took this silence as approval and pushed for a release.

In the future, I'll make sure to check with `rust-windowing`-specific platform experts before pushing for any kind of release.

> I get it, it's sometimes hard to not have a release while porting other crates in the ecosystem to the new thing, but using temporary `git` references (for example) allows the new API to settle for a while to get more eyes on it and work out the final kinks (for a critical crate like RWH).

This is actually the original path I took. See rust-windowing/winit#2744, rust-windowing/glutin#1582 and rust-windowing/softbuffer#82. Like I said above, my rationale was that I'd already gotten enough eyes on the system that it was ready to be published.

But, ah well, this is recoverable. Chances are, with #104 we're likely to have a breaking release in the future anyways.

rib:
Cool, thanks for the follow up @notgull 

Sorry that I'm again a bit slow with responding - I actually just started a new job this week, so been a bit busy :)  as it happens though I'll be responsible for lots of Android + Rust enabling though, so fingers crossed I'll be in a good position to actively help with this kind of thing.

Yeah I wouldn't call it a screw up either - it's an awkward corner of Android - and I'll even be ready to accept egg on my face at some point when we discover there is some weird safety issue here that we've all overlooked.

To the best of my understanding currently though I think the conclusion above is correct. `ANativeWindow` is taken out of our hands and referenced by graphics API surfaces, so even if there were a safety issue with touching an `ANativeWindow` outside of the create/destroy protocol (which I also haven't seen anything to suggest that's the case, if holding a reference) then I don't think the `Active` system would be enough to address that issue.

Btw, although I generally try to follow `r/rust` and `r/rust_gamedev` I missed your blog post unfortunately. I guess the challenge with aiming to get feedback via Reddit is that (even though there are _lots_ of people that read reddit) there are still _very_ few people actively involved with Rust + Android atm, so your best chance at getting feedback is to probably reach out to people directly.

notgull:
> I actually just started a new job this week, so been a bit busy :) as it happens though I'll be responsible for lots of Android + Rust enabling though, so fingers crossed I'll be in a good position to actively help with this kind of thing.

Ooo, sounds fun! Good luck!

> Yeah I wouldn't call it a screw up either - it's an awkward corner of Android - and I'll even be ready to accept egg on my face at some point when we discover there is some weird safety issue here that we've all overlooked.

Thanks. My gut still tells me that there's some kind of edge case here that we're forgetting. I feel like, every time a project touches Java, it gets ten times more complex for no reason.

> Btw, although I generally try to follow `r/rust` and `r/rust_gamedev` I missed your blog post unfortunately. I guess the challenge with aiming to get feedback via Reddit is that (even though there are _lots_ of people that read reddit) there are still _very_ few people actively involved with Rust + Android atm, so your best chance at getting feedback is to probably reach out to people directly.

I guess that's the curse of trying to do things that not many people are doing; there just aren't enough experts to go around.

> To the best of my understanding currently though I think the conclusion above is correct. `ANativeWindow` is taken out of our hands and referenced by graphics API surfaces, so even if there were a safety issue with touching an `ANativeWindow` outside of the create/destroy protocol (which I also haven't seen anything to suggest that's the case, if holding a reference) then I don't think the `Active` system would be enough to address that issue.

I'll take this as approval of the new system. Once @MarijnS95 signs off on this I'll go ahead and merge it.

notgull:
@MarijnS95 Poke, does this look good to you?

MarijnS95:
@notgull perhaps, I haven't had the time to look into anything over the Easter weekend, sorry for enjoying time off and slacking on this!

src/borrowed.rs
```diff
@@ -77,14 +45,15 @@ impl fmt::Debug for Active {
 /// On non-Android platforms, this is a ZST. On Android, this is a reference counted handle that
 /// keeps the application active while it is alive.
 #[derive(Clone)]
-pub struct ActiveHandle<'a>(imp::ActiveHandle<'a>);
+pub struct ActiveHandle<'a>(PhantomData<&'a UnsafeCell<()>>);
```
MarijnS95, line 48: Should the docs here be updated?

MarijnS95 (approved):
I'm not familiar with the new "borrowed" API of `raw-window-handle` and have just pointed out some generic findings on the diff. Overall the gain in removing this `Active` handling on Android is solid since the handle won't be pulled out from under you if holding a reference to it (which the NDK crate does for you; and graphics API's like Vulkan do too internally). The handle will just be "useless".

notgull:
Thank you!

PR #119, opened 2023-04-19
Add 3DS window handles

notgull:
This PR adds window handles for the Nintendo 3DS. Through libctru, the windows are represented as the top screen or the bottom screen, or as `GFX_TOP_SCREEN` and `GFX_BOTTOM_SCREEN`. This is represented as a `c_uint`.

### DIFF
--- a/src/horizon.rs
+++ b/src/horizon.rs
@@ -1,0 +1,47 @@
+use core::ffi::c_uint;
+
+/// Raw display handle for the 3DS.
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::HorizonDisplayHandle;
+/// let mut display_handle = HorizonDisplayHandle::empty();
+/// /* set fields */
+/// ```
+#[non_exhaustive]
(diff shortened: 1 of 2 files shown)


Lokathor:
If it's for ARMv6, do other ARM versions have other types? Should it be HorizonV6 or something?

Are the expected `screen` values just 0 or 1? or who decides that? if there's some other rust crate or 3ds doc we can link to we should consider that.

notgull:
> If it's for ARMv6, do other ARM versions have other types? Should it be HorizonV6 or something?

IIRC 3DS only used ARMv6, although I'd have to double check.

> Are the expected `screen` values just 0 or 1? or who decides that? if there's some other rust crate or 3ds doc we can link to we should consider that.

The screen values are set in [libctru](https://github.com/devkitPro/libctru/blob/4e25fb1d6c2ea124a9011c4b65f76f2968a9fb97/libctru/include/3ds/gfx.h#L24-L28).

notgull:
Needs more research, especially when it comes to the "right" way to interface with the 3DS API.

PR #124, opened 2023-06-10
Fix changelog

notgull:


Lokathor:
0.5.2 is released already and i didn't do the changelog at the time, so you'll need to check dates on PRs and determine which were merged before 0.5.2 and which parts should be still "unreleased" and set up for the 0.5.3

notgull:
- #110 and its sub-PRs (#115, #116, #117) were merged before 0.5.2
- #118 is not yet released/

Lokathor (approved):

PR #126, opened 2023-06-17
Remove Active from the API

notgull:
Since it's unnecessary and #122 is a breaking change anyways.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * **Breaking:** `HasRaw(Display/Window)Handle::raw_(display/window)_handle` returns a result indicating if fetching the window handle failed (#122).
-* Remove Android-specific platform differences (#118).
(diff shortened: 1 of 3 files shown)


Lokathor (approved):
I'm not an android expert but if we added this just for them and we don't need it anymore then it's fine to remove.

CHANGELOG.md
```diff
@@ -3,7 +3,7 @@
 ## Unreleased
 
 * **Breaking:** `HasRaw(Display/Window)Handle::raw_(display/window)_handle` returns a result indicating if fetching the window handle failed (#122).
-* Remove Android-specific platform differences (#118).
+* **Breaking:** Remove the `Active/ActiveHandle` types from the public API (#123).
```
MarijnS95, line 6:
```suggestion
* **Breaking:** Remove the `Active/ActiveHandle` types from the public API (#126).
```

MarijnS95 (approved):
Agreed on the idea and justification, haven't reviewed if this is all the active code/comments though but you know better :)

Issue #123, opened 2023-06-10
Would it make sense to have some kind of "view" or "renderable" handle?


madsmtm:
I'm only really familiar with macOS and iOS, but on those platforms it would make sense to split the window and the view handles in two.

That way, users could make some sort of interface in the native toolkit, and just provide the `NSView` handle to `wgpu` to render into. With the current design, we're assuming that the whole window's content is what should be rendered into.

I think this is roughly how it already works on web, e.g. you can lay out the page in a certain way, and then just create a canvas that you pass on to the renderer to do its thing. (Though arguably `winit` shouldn't be creating a canvas, perhaps instead it should provide a `div` that the graphics library itself can insert the canvas into?)

Would this sort of thing work on other platforms?

MarijnS95:
> I think this is roughly how it already works on web

And on android, where `NativeWindow` is just a silly name for the [native wrapper around a `Surface`](https://developer.android.com/ndk/reference/group/a-native-window#summary) and could be any part of the view layout, not necessarily the whole "window surface".

https://github.com/MarijnS95/AndroidNativeSurface

madsmtm:
Basically, I want better cooperation with the existing layout systems (if they exist).

I'm especially unsure what the best approach is for building a "native" user interface on Linux? GTK? Or would it be best to do it "raw", and provide something like `struct XcbViewHandle { visual_id: u32, position: (u32, u32), size: (u32, u32) }`?

kchibisov:
Wayland's wl_surface can be used for anything as well, but that's how the handle is done.

I wonder whether on macOS you don't need to have a reference to a window to perform some requests on it? Also, the user could simply omit the window, fields are not required to be set, so setting just NSView is absolutely fine.

> and provide something like struct XcbViewHandle { visual_id: u32, position: (u32, u32), size: (u32, u32) }?

visual has nothing to do with windows on X11, it's a pixel format. Everything is a XID, which is window as it's said in the structs.

madsmtm:
The window field is entirely redundant on macOS, it can always be gotten from the view.

So maybe it would actually make sense to rename the `[Raw]WindowHandle` to `[Raw]ViewHandle` (or similar)?

MarijnS95:
> to `[Raw]ViewHandle` (or similar)?

The common name for this thing on e.g. Android and in graphics APIs like (E)GL and Vulkan is a "surface", I'd name it that.

[...]

PR #129, opened 2023-06-24
Remove window handles from Apple handles

madsmtm:
The window can easily be fetched from the view, and having it leads to confusion over which place you should be e.g. taking the drawing dimensions (always the view).

Part of https://github.com/rust-windowing/raw-window-handle/issues/123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * **Breaking:** `HasRaw(Display/Window)Handle::raw_(display/window)_handle` returns a result indicating if fetching the window handle failed (#122).
 * **Breaking:** Remove the `Active/ActiveHandle` types from the public API (#126).
+* **Breaking:** Remove `AppKitWindowHandle::ns_window` and `UiKitWindowHandle::ui_window` since they can be retrived from the view (#129).
 
 ## 0.5.2 (2023-03-31)
 
--- a/src/appkit.rs
+++ b/src/appkit.rs
@@ -30,17 +30,13 @@ impl AppKitDisplayHandle {
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct AppKitWindowHandle {
-    /// A pointer to an `NSWindow` object.
-    pub ns_window: *mut c_void,
     /// A pointer to an `NSView` object.
     pub ns_view: *mut c_void,
-    // TODO: WHAT ABOUT ns_window_controller and ns_view_controller?
 }
 
 impl AppKitWindowHandle {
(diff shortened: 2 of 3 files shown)


notgull (commented):
I think there's precedent for keeping derivable information in the window handle itself; the X11 window handles contain the visual ID despite the fact that you can figure out the visual from the window. This could be fixed by just adjusting the order/documentation to make it clearer that the view takes precedence.

Although it would be *fine*, the reason why I'm opposed is because I was planning on doing the semver trick by re-exporting the `RawWindowHandle` from v0.6 in v0.5 to avoid breaking that part of the API. If there's a breaking change in the window handle here that would become impossible or at least significantly less effective.

That being said it's macOS, you definitely know best in this area. Your call.

kchibisov:
I think the reason we have them is to be able to pass opaque pointers around and don't pull dependencies to take `ns_window` or `x11_visual`, it's true that you may not set them, but it could make things more complicated. We do use `ns_window` in glutin for example, though, we already have `objc` pulled.

madsmtm:
The problem is that the documentation says that an implementation of the trait is allowed to only fill in the view, and that the user of the trait, if the window is not present, must then figure it out from the view. So a fully conforming user must pull in `objc2` to get the window anyhow.

Another point against providing redundant information is that you have to ensure that they are in sync.

madsmtm:
I don't think this prevents the semver trick? It should go something like:
```rust
// In version 0.5.4
impl<T: v6::HasRawWindowHandle> HasRawWindowHandle for T {
        match ... {
            v6::AppKitHandle(handle) => {
                let new_handle = AppKitHandle::empty();
                new_handle.ns_view = handle.ns_view;
                new_handle
            }
            ...
        }
    }
}
```

Lokathor:
The current fields are that way because SDL2 provides the window and does not give the view.

madsmtm:
Perhaps any SDL2 Rust crate that uses `raw-window-handle` should be changed to output the view then? Since that is what anything that wants to render will have to do anyway. Should just be a call to `[window contentView]`.

notgull:
> Perhaps any SDL2 Rust crate that uses `raw-window-handle` should be changed to output the view then? Since that is what anything that wants to render will have to do anyway. Should just be a call to `[window contentView]`.

This sounds like a sane solution. @Lokathor would this work for SDL2 crates?

LGTM otherwise

Lokathor:
Yeah I mean I guess that sounds fine? I'm not an apple user really.

There's some changelog conflicts that have cropped up that need to be fixed it seems

notgull (approved):

kchibisov (approved):

Issue #125, opened 2023-06-17
`v0.6` release planning


madsmtm:
What do we need to fix before the next major release?

There are requests for us doing the server trick again, which I'll be up for (trying to do), so mostly opening this issue to keep track of that.

notgull:
I wanted to get rid of the `Active` type since we're doing a breaking change anyways.

notgull:
> I wanted to get rid of the `Active` type since we're doing a breaking change anyways.

Did this in #126. Anything else that we need to think about?

madsmtm:
Hmm, I think the blanket implementations for `&T` and so on are going to be a problem for us, since we effectively want to do a blanket impl for `T: new::HasWindowHandle`.

Still considering this, but perhaps we could do the opposite? Something like `T: old::HasWindowHandle`, and then prepare for v7 where it would not be a problem?

Otherwise, we'll have to encourage either producer crates or consumer crates to support v5 for at least a release cycle or two.

madsmtm:
Extra possibility: We could consider removing the blanket impls in v6, to allow v7 to properly use the semver trick again.

madsmtm:
Blocked by (at least discussion of):
- [ ] https://github.com/rust-windowing/raw-window-handle/pull/136
- [ ] https://github.com/rust-windowing/raw-window-handle/issues/135

madsmtm:
v0.6 has been released

PR #131, opened 2023-07-02
Make raw handle traits safe to implement

notgull:
Here's a thought experiment that we should at least *consider* before releasing v0.6 (#125). Since we now have borrowed display/window handles, with conditions on them enforced via their `borrow_raw` methods, what if we made the raw traits safe to implement? They're unsafe to implement now, as contrasted with `AsRawFd` in the standard library, which are safe.

The primary downside would be the introduction of a footgun where users try to use `HasRawWindowHandle` in their APIs are they do now, unaware that they can return undefined window handles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * **Breaking:** `HasRaw(Display/Window)Handle::raw_(display/window)_handle` returns a result indicating if fetching the window handle failed (#122).
 * **Breaking:** Remove the `Active/ActiveHandle` types from the public API (#126).
+* **Breaking:** It is now safe to implement `HasRaw(Display/Window)Handle` (#130).
 
 ## 0.5.2 (2023-03-31)
 
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -26,8 +26,13 @@ use crate::{
 ///
 /// # Safety
 ///
-/// The safety requirements of [`HasRawDisplayHandle`] apply here as well. To reiterate, the
-/// [`DisplayHandle`] must contain a valid window handle for its lifetime.
+/// Users can safely assume that non-`null`/`0` fields are valid handles, and it is up to the
+/// implementer of this trait to ensure that condition is upheld.
+///
+/// Despite that qualification, implementers should still make a best-effort attempt to fill in all
+/// available fields. If an implementation doesn't, and a downstream user needs the field, it should
+/// try to derive the field from other fields the implementer *does* provide via whatever methods the
+/// platform provides.
 ///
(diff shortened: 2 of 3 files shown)


i509VCB:
I'm skeptical there is a benefit to this. `HasRawDisplayHandle` and `HasRawWindowHandle` imply that the implementer either owns the underlying handle(s) or borrows the handles (which would require following the contract of the API).

Regarding the comparison to `AsRawFd` I feel HasRawWindowHandle differs in the semantics. If an fd is closed and you use the fd somewhere, you'll get some errno or `io::Error` return value. Whereas with raw window/display handles you are likely passing around a pointer which on some platforms could be dangling if the owner of the handles is freed.

ids1024:
> They're unsafe to implement now, as contrasted with AsRawFd in the standard library, which are safe.

The safety, and possibly the existence, of this trait is arguably just a relic of it predating IO safety. And the fact that can't be changed.

Is `HasRawWindowHandle` handle still needed? Or would it be better for all implementers to just implement `HasWindowHandle`, and get a raw window handle from that, as needed? Is there any reason to implement `HasRawWindowHandle` but not `HasWindowHandle`?

>If an fd is closed and you use the fd somewhere, you'll get some errno or io::Error return value. 

Not necessarily. The fd may have been re-used, and may not point to a different file description. Which may be mapped into memory, so you may cause undefined behavior by writing to it. 

So you *might* get an IO error, but that can also be unsound, and that is basically the reason for the IO safety rules and types.

i509VCB:
If I recall the end goal of the regular WindowHandle type is to make the whole process of getting a softbuffer/gl context/etc safe. However that would need a lifetime on context types? So a Display<'a> for example. But some use cases you pretty much can't have a lifetime, (wayland-rs Dispatch: 'static is very guilty of this). I guess you could transmute a Display<'a> to a Display<'static> but that would require writing transmutes and you are heavily discouraged from doing that plus it's a UB nightmare waiting to happen.

I guess ouroboros could be a workaround to the "context needs to be owned with the display" scenario but using ouroboros feels unclean.

notgull:
I agree with the rationale presented here, so I’ll close this PR.

The intended solution to the problem of a borrowed window handle having a lifetime parameter is to pass around the `T: HasWindowHandle` itself. I've applied this pattern to my conversions so far, see https://github.com/rust-windowing/softbuffer/pull/132, https://github.com/gfx-rs/wgpu/pull/3932 and https://github.com/rust-windowing/glutin/pull/1582. This way you can use `WindowHandle<'a>`, `Rc<Window>` and `Arc<Window>` in all of these places without any issues.

`ouroboros` is sketchy, I agree.

PR #133, opened 2023-07-16
Implement traits for &mut T

notgull:
As the traits are implemented for `&T` where `T` implements the traits, I'm not sure why they can't also be implemented for `&mut T`.


Lokathor:
They definitely can be. I suspect this wasn't done because of how rarely you'd have a &mut T on them and how that would coerce into &T anyway, but sometimes coercion doesn't kick in right so we might as well have these impls

Lokathor (approved):

PR #134, opened 2023-07-17
Add wasm-bindgen v0.2 handles for web_sys::HtmlCanvasElement and web_sys::OffscreenCanvas

### LINKED ISSUE #102: Feature for `web_sys::HtmlCanvasElement` or `web_sys::OffscreenCanvas`
Following from https://github.com/rust-windowing/winit/issues/2259, the `WebWindowHandle.id` is a flaky way to share canvas references between crates. As discussed in that thread, synchronizing the IDs is challenging (leading to duplicates) and it isn't possible to query canvas elements in a Shadow DOM or offscreen canvases at all.

What consumers need is a way to pass the canvas reference through `HasRawWindowHandle`. AFAICT, that can be done with either an enum with variants for `web_sys::HtmlCanvasElement` and `web_sys::OffscreenCanvas`, or with `wasm_bindgen::JsValue`. It looks like another option is the `FromWasmAbi` and `IntoWasmAbi` traits provided by `wasm_bindgen` if the reference needs to cross the WASM ABI boundary for some reason.

I think the best way forward in the short-term is a feature to enable the use of one of these reference types in addition to the `id: u32`. It is preferable to replace the id entirely but may not be realistic without a deprecation period. The feature also mitigates a potential downside that it adds a dependency on `web_sys` and/or `wasm_bindgen`.

Open to discussion and other alternatives, but the current design appears to be a significant shortcoming in web environments.

notgull:
Closes #102 by adding two new window handles. One represents a `web_sys::HtmlCanvasElement` registered into `wasm-bindgen` through its index and the other represents a `web_sys::OffscreenCanvas` in the same way. This allows the handles to contain arbitrary canvases without needing to rely on the `id` hack.

The main issue with this PR is that the underlying representation for `wasm-bindgen` objects might change to something aside from the current index system. This was raised in https://github.com/rustwasm/wasm-bindgen/issues/1766. However, the index system has remained in place for five years and according to @daxpedda it isn't going anywhere soon. In addition a new system would likely require a breaking change. Therefore we should be fine with this system for the foreseeable future.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,14 +15,8 @@ rust-version = "1.64"
 alloc = []
 std = ["alloc"]
 
-# Unstable feature to allow the use of `wasm-bindgen` in the `wasm` target.
-# 
-# This is radioactive semver-wise. Do not rely on this in stable code. Any code enabled by this
-# feature can be changed or removed in a breaking way at any time. In fact, this feature could
-# be removed entirely tomorrow.
-# 
-# The point being: rely on this at your own peril.
-unstable_web_handles_wbg_02 = ["wasm-bindgen", "std"]
+# Allow conversion methods to/from WASM types using `wasm-bindgen` v0.2.
+wasm-bindgen-0-2 = ["wasm-bindgen", "std"]
 
 [target.'cfg(target_family = "wasm")'.dependencies.wasm-bindgen]
 version = "0.2.87"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ std = ["alloc"]
 # be removed entirely tomorrow.
 # 
 # The point being: rely on this at your own peril.
-unstable_web_handles = ["wasm-bindgen", "std"]
+unstable_web_handles_wbg_02 = ["wasm-bindgen", "std"]
 
 [target.'cfg(target_family = "wasm")'.dependencies.wasm-bindgen]
 version = "0.2.87"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,21 @@ rust-version = "1.64"
 alloc = []
 std = ["alloc"]
 
+# Unstable feature to allow the use of `wasm-bindgen` in the `wasm` target.
+# 
+# This is radioactive semver-wise. Do not rely on this in stable code. Any code enabled by this
+# feature can be changed or removed in a breaking way at any time. In fact, this feature could
+# be removed entirely tomorrow.
+# 
+# The point being: rely on this at your own peril.
+unstable_web_handles = ["wasm-bindgen", "std"]
+
+[target.'cfg(target_family = "wasm")'.dependencies.wasm-bindgen]
+version = "0.2.87"
+default-features = false
+features = ["std"]
+optional = true
+
 [package.metadata.docs.rs]
 all-features = true
 rustdoc-args = ["--cfg", "docsrs"]
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -84,7 +84,7 @@ impl<H: HasDisplayHandle + ?Sized> HasDisplayHandle for alloc::sync::Arc<H> {
 ///
 /// Get the underlying raw display handle with the [`HasRawDisplayHandle`] trait.
 #[repr(transparent)]
-#[derive(PartialEq, Eq, Hash, Copy, Clone)]
+#[derive(PartialEq, Clone)]
 pub struct DisplayHandle<'a> {
     raw: RawDisplayHandle,
     _marker: PhantomData<&'a *const ()>,
@@ -112,13 +112,13 @@ impl<'a> DisplayHandle<'a> {
 
 unsafe impl HasRawDisplayHandle for DisplayHandle<'_> {
     fn raw_display_handle(&self) -> Result<RawDisplayHandle, HandleError> {
-        Ok(self.raw)
+        Ok(self.raw.clone())
     }
 }
 
 impl<'a> HasDisplayHandle for DisplayHandle<'a> {
     fn display_handle(&self) -> Result<DisplayHandle<'_>, HandleError> {
-        Ok(*self)
+        Ok(self.clone())
     }
 }
 
@@ -209,7 +209,7 @@ impl<H: HasWindowHandle + ?Sized> HasWindowHandle for alloc::sync::Arc<H> {
 ///
 /// This handle is guaranteed to be safe and valid. Get the underlying raw window handle with the
 /// [`HasRawWindowHandle`] trait.
-#[derive(PartialEq, Eq, Hash, Clone)]
+#[derive(PartialEq, Clone)]
 pub struct WindowHandle<'a> {
     raw: RawWindowHandle,
     _marker: PhantomData<&'a *const ()>,
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -84,7 +84,7 @@ impl<H: HasDisplayHandle + ?Sized> HasDisplayHandle for alloc::sync::Arc<H> {
 ///
 /// Get the underlying raw display handle with the [`HasRawDisplayHandle`] trait.
 #[repr(transparent)]
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Eq, Hash, Clone)]
 pub struct DisplayHandle<'a> {
     raw: RawDisplayHandle,
     _marker: PhantomData<&'a *const ()>,
@@ -112,7 +112,7 @@ impl<'a> DisplayHandle<'a> {
 
 unsafe impl HasRawDisplayHandle for DisplayHandle<'_> {
     fn raw_display_handle(&self) -> Result<RawDisplayHandle, HandleError> {
-        Ok(self.raw.clone())
+        Ok(self.raw)
     }
 }
 
@@ -209,7 +209,7 @@ impl<H: HasWindowHandle + ?Sized> HasWindowHandle for alloc::sync::Arc<H> {
 ///
 /// This handle is guaranteed to be safe and valid. Get the underlying raw window handle with the
 /// [`HasRawWindowHandle`] trait.
-#[derive(PartialEq, Clone)]
+#[derive(PartialEq, Eq, Hash, Clone)]
 pub struct WindowHandle<'a> {
     raw: RawWindowHandle,
     _marker: PhantomData<&'a *const ()>,
@@ -237,7 +237,7 @@ impl<'a> WindowHandle<'a> {
 
 unsafe impl HasRawWindowHandle for WindowHandle<'_> {
     fn raw_window_handle(&self) -> Result<RawWindowHandle, HandleError> {
-        Ok(self.raw.clone())
+        Ok(self.raw)
     }
 }
 
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -84,7 +84,7 @@ impl<H: HasDisplayHandle + ?Sized> HasDisplayHandle for alloc::sync::Arc<H> {
 ///
 /// Get the underlying raw display handle with the [`HasRawDisplayHandle`] trait.
 #[repr(transparent)]
-#[derive(PartialEq, Eq, Hash, Clone)]
+#[derive(PartialEq, Eq, Hash, Copy, Clone)]
 pub struct DisplayHandle<'a> {
     raw: RawDisplayHandle,
     _marker: PhantomData<&'a *const ()>,
@@ -118,7 +118,7 @@ unsafe impl HasRawDisplayHandle for DisplayHandle<'_> {
 
 impl<'a> HasDisplayHandle for DisplayHandle<'a> {
     fn display_handle(&self) -> Result<DisplayHandle<'_>, HandleError> {
-        Ok(self.clone())
+        Ok(*self)
     }
 }
 
@@ -209,7 +209,7 @@ impl<H: HasWindowHandle + ?Sized> HasWindowHandle for alloc::sync::Arc<H> {
 ///
 /// This handle is guaranteed to be safe and valid. Get the underlying raw window handle with the
 /// [`HasRawWindowHandle`] trait.
-#[derive(PartialEq, Eq, Hash, Clone)]
+#[derive(PartialEq, Eq, Hash, Copy, Clone)]
 pub struct WindowHandle<'a> {
     raw: RawWindowHandle,
     _marker: PhantomData<&'a *const ()>,
@@ -243,7 +243,7 @@ unsafe impl HasRawWindowHandle for WindowHandle<'_> {
 
 impl HasWindowHandle for WindowHandle<'_> {
     fn window_handle(&self) -> Result<Self, HandleError> {
-        Ok(self.clone())
+        Ok(*self)
     }
 }
 
--- a/src/web.rs
+++ b/src/web.rs
@@ -58,16 +58,15 @@ impl WebWindowHandle {
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct WebCanvasWindowHandle {
-    /// The pointer to the [`JsValue`] of an [`HtmlCanvasElement`].
+    /// A pointer to the [`JsValue`] of an [`HtmlCanvasElement`].
     ///
-    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
     /// [`JsValue`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html
+    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
     pub obj: NonNull<c_void>,
 }
 
 impl WebCanvasWindowHandle {
-    /// Create a new handle to a pointer to [`HtmlCanvasElement`].
+    /// Create a new handle from a pointer to [`HtmlCanvasElement`].
     ///
     /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
     ///
@@ -97,53 +96,41 @@ impl WebCanvasWindowHandle {
 )]
 /// These implementations are only available when `wasm-bindgen-0-2` is enabled.
 impl WebCanvasWindowHandle {
-    /// Create a new `Wbg02CanvasWindowHandle` from a [`wasm-bindgen`] object.
+    /// Create a new `WebCanvasWindowHandle` from a [`wasm_bindgen::JsValue`].
     ///
-    /// This function is unstable. Its signature may be changed or even removed outright without a
-    /// breaking version change.
-    ///
-    /// # Safety
-    ///
-    /// The [`JsValue`] must refer to an [`HtmlCanvasElement`], and the lifetime must be longer than
-    /// the `Wbg02CanvasWindowHandle` lives for.
-    ///
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// The `JsValue` should refer to a `HtmlCanvasElement`, and the lifetime
+    /// of the value should be at least as long as the lifetime of this.
     pub fn from_wasm_bindgen_0_2(js_value: &wasm_bindgen::JsValue) -> Self {
         Self::new(NonNull::from(js_value).cast())
     }
 
-    /// Convert to the underlying [`wasm-bindgen`] index.
-    ///
-    /// This function is unstable. Its signature may be changed or even removed outright without a
-    /// breaking version change.
+    /// Convert to the underlying [`wasm_bindgen::JsValue`].
     ///
     /// # Safety
     ///
-    /// The lifetime from the `from_wasm_bindgen_0_2` function must still be valid, and the
-    /// underlying pointer must still be a [`wasm_bindgen`] object.
-    ///
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// The inner pointer must be valid. This is ensured if this handle was
+    /// borrowed from [`WindowHandle`][crate::WindowHandle].
     pub unsafe fn as_wasm_bindgen_0_2(&self) -> &wasm_bindgen::JsValue {
-        self.obj.cast().as_ref()
+        unsafe { self.obj.cast().as_ref() }
     }
 }
 
-/// Raw window handle for a Web offscreen canvas registered via [`wasm-bindgen`].
+/// Raw window handle for a Web offscreen canvas registered via
+/// [`wasm-bindgen`].
 ///
 /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct WebOffscreenCanvasWindowHandle {
-    /// The pointer to the [`JsValue`] of an [`OffscreenElement`].
+    /// A pointer to the [`JsValue`] of an [`OffscreenCanvas`].
     ///
-    /// [`OffscreenElement`]: https://docs.rs/web-sys/latest/web_sys/struct.OffscreenElement.html
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
     /// [`JsValue`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html
+    /// [`OffscreenCanvas`]: https://docs.rs/web-sys/latest/web_sys/struct.OffscreenCanvas.html
     pub obj: NonNull<c_void>,
 }
 
 impl WebOffscreenCanvasWindowHandle {
-    /// Create a new handle to a pointer to an [`OffscreenCanvas`].
+    /// Create a new handle from a pointer to an [`OffscreenCanvas`].
     ///
     /// [`OffscreenCanvas`]: https://docs.rs/web-sys/latest/web_sys/struct.OffscreenCanvas.html
     ///
@@ -173,33 +160,22 @@ impl WebOffscreenCanvasWindowHandle {
 )]
 /// These implementations are only available when `wasm-bindgen-0-2` is enabled.
 impl WebOffscreenCanvasWindowHandle {
-    /// Create a new `Wbg02OffscreenCanvasWindowHandle` from a [`wasm-bindgen`] object.
+    /// Create a new `WebOffscreenCanvasWindowHandle` from a
+    /// [`wasm_bindgen::JsValue`].
     ///
-    /// This function is unstable. Its signature may be changed or even removed outright without a
-    /// breaking version change.
-    ///
-    /// # Safety
-    ///
-    /// The [`JsValue`] must refer to an [`OffscreenCanvas`], and the lifetime must be longer than
-    /// the `Wbg02OffscreenCanvasWindowHandle` lives for.
-    ///
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// The `JsValue` should refer to a `HtmlCanvasElement`, and the lifetime
+    /// of the value should be at least as long as the lifetime of this.
     pub fn from_wasm_bindgen_0_2(js_value: &wasm_bindgen::JsValue) -> Self {
         Self::new(NonNull::from(js_value).cast())
     }
 
-    /// Convert to the underlying [`wasm-bindgen`] index.
-    ///
-    /// This function is unstable. Its signature may be changed or even removed outright without a
-    /// breaking version change.
+    /// Convert to the underlying [`wasm_bindgen::JsValue`].
     ///
     /// # Safety
     ///
-    /// The lifetime from the `from_wasm_bindgen_0_2` function must still be valid, and the
-    /// underlying pointer must still be a [`wasm_bindgen`] object.
-    ///
-    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// The inner pointer must be valid. This is ensured if this handle was
+    /// borrowed from [`WindowHandle`][crate::WindowHandle].
     pub unsafe fn as_wasm_bindgen_0_2(&self) -> &wasm_bindgen::JsValue {
-        self.obj.cast().as_ref()
+        unsafe { self.obj.cast().as_ref() }
     }
 }
--- a/src/web.rs
+++ b/src/web.rs
@@ -48,3 +48,67 @@ impl WebWindowHandle {
         Self { id }
     }
 }
+
+/// Raw window handle for a Web canvas registered via [`wasm-bindgen`].
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::Wbg02CanvasWindowHandle;
+/// let mut window_handle = Wbg02CanvasWindowHandle::empty();
+/// /* set fields */
+/// ```
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Wbg02CanvasWindowHandle {
+    /// The index of the canvas element in the [`wasm-bindgen`] table.
+    ///
+    /// If this index if non-zero, it is implied that it represents an [`HtmlCanvasElement`]
+    /// that is registered in the [`wasm-bindgen`] table. It can be converted to and from the
+    /// [`HtmlCanvasElement`] using [`wasm-bindgen`]'s [`FromWasmAbi`] and [`IntoWasmAbi`] traits.
+    ///
+    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
+    /// [`FromWasmAbi`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/trait.FromWasmAbi.html
+    /// [`IntoWasmAbi`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/trait.IntoWasmAbi.html
+    pub idx: u32,
+}
+
+impl Wbg02CanvasWindowHandle {
+    pub fn empty() -> Self {
+        Self { idx: 0 }
+    }
+}
+
+/// Raw window handle for a Web offscreen canvas registered via [`wasm-bindgen`].
+///
+/// ## Construction
+/// ```
+/// # use raw_window_handle::Wbg02OffscreenCanvasWindowHandle;
+/// let mut window_handle = Wbg02OffscreenCanvasWindowHandle::empty();
+/// /* set fields */
+/// ```
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Wbg02OffscreenCanvasWindowHandle {
+    /// The index of the canvas element in the [`wasm-bindgen`] table.
+    ///
+    /// If this index if non-zero, it is implied that it represents an [`OffscreenCanvas`]
+    /// that is registered in the [`wasm-bindgen`] table. It can be converted to and from the
+    /// [`OffscreenCanvas`] using [`wasm-bindgen`]'s [`FromWasmAbi`] and [`IntoWasmAbi`] traits.
+    ///
+    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// [`OffscreenCanvas`]: https://docs.rs/web-sys/latest/web_sys/struct.OffscreenCanvas.html
+    /// [`FromWasmAbi`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/trait.FromWasmAbi.html
+    /// [`IntoWasmAbi`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/trait.IntoWasmAbi.html
+    pub idx: u32,
+}
+
+impl Wbg02OffscreenCanvasWindowHandle {
+    pub fn empty() -> Self {
+        Self { idx: 0 }
+    }
+}
--- a/src/web.rs
+++ b/src/web.rs
@@ -1,4 +1,5 @@
-use core::fmt;
+use core::ffi::c_void;
+use core::ptr::NonNull;
 
 /// Raw display handle for the Web.
 #[non_exhaustive]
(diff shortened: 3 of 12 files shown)


grovesNL:
Could we expose the regular web-sys types (`HtmlCanvasElement` and `OffscreenCanvas`) instead? `FromWasmAbi` and `IntoWasmAbi` transfer ownership of the JavaScript object, so it's really easy to get this wrong and drop the window, ending up with a use-after-free by accident.

notgull:
> Could we expose the regular web-sys types (`HtmlCanvasElement` and `OffscreenCanvas`) instead? `FromWasmAbi` and `IntoWasmAbi` transfer ownership of the JavaScript object, so it's really easy to get this wrong and drop the window, ending up with a use-after-free by accident.

This would require importing the `web-sys` crate, which is bad for a couple of reasons.

- It requires us to have `cfg` guards on the window handle itself, which is something I'm trying to avoid (#63)
- `web-sys` is a very heavy dependency for a crate that's supposed to be lean and mean.

grovesNL:
For what it's worth wgpu tried to use `WasmAbi` functions for a similar use case and ended up having to back them out because of the ownership issues (https://github.com/gfx-rs/wgpu/issues/3430 has some more context if you're interested).

> It requires us to have `cfg` guards on the window handle itself

Could we avoid this by making the web-sys handles type aliases or opaque types that use `()` on other targets?

> `web-sys` is a very heavy dependency for a crate that's supposed to be lean and mean.

Agreed but `web-sys` will most likely be in use already for people using this target, and we could still use cfg/feature guards if anyone would like to opt-out.

notgull:
> For what it's worth wgpu tried to use `WasmAbi` functions for a similar use case and ended up having to back them out because of the ownership issues ([gfx-rs/wgpu#3430](https://github.com/gfx-rs/wgpu/issues/3430) has some more context if you're interested).

We could likely solve this on our end by making the `FromWasmAbi` results `ManuallyDrop`. This is how it's done in other places where raw pointers are used like this, like in `Waker`s.

> Could we avoid this by making the web-sys handles type aliases or opaque types that use `()` on other targets?

This adds a layer of `cfg` complexity I'm uncomfortable having.

grovesNL:
At least anecdotally based on my experience removing the use of `WasmAbi` in wgpu, I wouldn't feel confident accepting these handles and converting them back into `web_sys` types. For example, if one raw window handle is reused to create two `HtmlCanvasElement` from the `u32` representation then this would break in a subtle way.

notgull:
> At least anecdotally based on my experience removing the use of `WasmAbi` in wgpu, I wouldn't feel confident accepting these handles and converting them back into `web_sys` types. For example, if one raw window handle is reused to create two `HtmlCanvasElement` from the `u32` representation then this would break in a subtle way.

Unless there's a subtlety I'm missing it seems like just wrapping the result of `FromWasmAbi` in `ManuallyDrop` should be enough to prevent this double-drop from occurring.

grovesNL:
My understanding is that it goes something like this because of the ownership transfer that happens when you go through `IntoWasmAbi`/`FromWasmAbi`:
- the owner of `HtmlCanvasElement` (e.g., winit) calls `into_abi` and gets a value like `10u32` and puts it into a raw window handle
- then the raw window handle is handed to another crate (e.g., wgpu) and `10u32` is converted back to `HtmlCanvasElement` (optionally wrapped with `ManuallyDrop`) and it's used for some initial setup code, eventually dropped
- then the raw window handle is used again (e.g., same crate as before or another) but `10u32` isn't actually valid to be used for the next `from_abi` call

notgull:
> - then the raw window handle is handed to another crate (e.g., wgpu) and `10u32` is converted back to `HtmlCanvasElement` (optionally wrapped with `ManuallyDrop`) and it's used for some initial setup code, **eventually dropped**

Emphasis mine. As long as the canvas element is never dropped by the other crate there isn't a double drop. It can be either forgotten or wrapped in `ManuallyDrop` as I did above. As long as the original window source is kept alive you can be sure that the canvas reference is valid, and this can now be done thanks to borrowed window handles.

grovesNL:
In my example, wgpu wouldn't need the handle anymore so wgpu would drop it, but I thought this would invalidate the ABI handle for future reuse (e.g. passing it to wgpu or another crate). Is my understanding wrong?

daxpedda:
> In my example, wgpu wouldn't need the handle anymore so wgpu would drop it, but I thought this would invalidate the ABI handle for future reuse (e.g. passing it to wgpu or another crate).

I didn't follow the exact discussion so far, but if you drop the `JsValue`, any "pointers" to it will become invalid indeed. A trait implementer could avoid this by cloning the `JsValue` internally and return the "pointer" to that.

notgull:
To sum up what was discussed in the Matrix room: the idea is that ownership isn't really passed through the `idx` value to `wgpu`/`glutin`/`softbuffer`/etc. The `winit::Window` or whatever is what really owns the `Canvas`; when the `Window` is dropped the `Canvas` is dropped.

The renderer, then, only owns a "reference" to that `Canvas`. It would call `FromWasmAbi` to get a `Canvas`, place that in a `ManuallyDrop`, and then *never drop it*. Semantically it's similar to a borrow. Once the renderer is dropped, the `Canvas` isn't dropped with it; it's just left for the windowing system to drop it.

In chat this was compared to `Arc::from_ptr()` and `Arc::into_ptr()`. This strategy is actually used somewhat frequently in cases where pointers are being passed around. See [this code](https://github.com/smol-rs/waker-fn/blob/6b47138087c81d53590a290e54e9092a9c195a03/src/lib.rs#L55-L58) from `smol` for an example.

@daxpedda Does this implementation look right? As you're the resident expert on `wasm-bindgen` I'll defer to you on this before I merge it.

daxpedda:
As far as I understood the explained implementation, it is correct. It would all be unsafe though until you somehow introduce a lifetime that binds it to the owner of `Canvas`.

notgull:
Yes, as @daxpedda said the "thing can be double dropped if you aren't careful" problem would not be unique to these handles. Pretty much every resource contained in these window handles have resources associated with them that can't be dropped twice. It's up the the windowing system to own it and the rendering system to not drop it.

grovesNL:
> I don't know much about RWH, but my impression so far was that this is the same issue other backends have as well.

That's true, although temporary canvases are a common pattern for wasm projects (in contrast to native windows which are typically long-lived), especially projects that aren't  using a windowing system like winit. Those projects won't be able to use raw-window-handle using this approach without leaking. This would work well if we used web-sys types instead.

notgull:
> That's true, although temporary canvases are a common pattern for wasm projects (in contrast to native windows which are typically long-lived), especially projects that aren't  using a windowing system like winit. Those projects won't be able to use raw-window-handle using this approach without leaking. 

Using the new borrowed handle system it should be possible to parameterize rendering types by the object managing the canvas. This should make it easy from a user perspective to write a wrapper around temporary canvases to work in this manner.

> This would work well if we used web-sys types instead.

One of the main sticking points for `raw-window-handle` is its lack of `cfg` guards and dependencies. I find the current solution, as hacky as it may be, preferable to going back on both of these points.

notgull:
Given that there have been no comments on this thread for a while, and that the Web maintainer for `winit` has signed off on this, it seems to me that there are no more open issues and that this can be merged.

I'll give it a week before I do just in case anyone else has any comments.

grovesNL:
> Using the new borrowed handle system it should be possible to parameterize rendering types by the object managing the canvas. This should make it easy from a user perspective to write a wrapper around temporary canvases to work in this manner.

In this case you'd want the rendering types to have a strong reference to the temporary canvas in this case, so the rendering types would effectively have to become the owner.

> One of the main sticking points for `raw-window-handle` is its lack of `cfg` guards and dependencies. I find the current solution, as hacky as it may be, preferable to going back on both of these points.

1. I think a `cfg` guard to alias web types internally is different than the concerns raised in #63 The point is that you have to be on the wasm target in order to use web-sys types anyway, so these are already guarded at some level (either at the crate level or wasm-specific paths in on both sides of handle creation/usage). In other words, downstream crates already have to guard these paths, so it shouldn't affect the ergonomics negatively as raised in #63

2. It's not possible to create or use these handles without web-sys already being somewhere in the dependency tree on this target, so while it's a nice goal to have as few dependencies as possible, it doesn't add dependencies if you actually want to use the handle. Almost any crate targeting the web/wasm will already have web-sys as a dependency.

I appreciate trying to be consistent in exposing these handles, and we can proceed with this if people believe it's the best path forward. I'd just like to reiterate that using these ABI handles is:
- unsafe and error-prone because it relies on the under-documented ABI conversion features of wasm-bindgen which in turn makes the ownership story unclear and easy to accidentally cause a use-after-free or leak canvases
- problematic trying to support common web use cases with a temporary canvas being handed off to the renderer

It would also be unfortunate to add this complexity back into wgpu after recently removing it because of the ownership/borrowing/refcounting challenges the ABI functions caused.

The alternative of wrapping web-sys types is what downstream crates would naturally expect, and these types are what people already pass today (if they're not going through raw-window-handle). I feel like the most pragmatic solution is to wrap the web-sys types and having a target-gated dependency, even though it's not consistent with how other handles are exposed<sup>1</sup>.

<sup>1</sup> Although I'd argue other handles would also try to use something better than `*mut c_void` if platforms/libraries could've historically agreed on a handle type that was stronger than that, like the handles web-sys provides in this case.

madsmtm:
I'll propose another solution: Expose the `JsValue` only through `cfg`-gated methods, and enable the methods based on a feature flag (instead of a specific platform).

Something like the following (showing an extension to the hypothetical future of a `web-sys v0.4`).
```toml
# Cargo.toml
[dependencies]
web-sys-0-3 = { package = "web-sys", version = "0.3" }
web-sys-0-4 = { package = "web-sys", version = "0.4" }
```

```rust
pub struct Wbg02CanvasWindowHandle(Inner);

// Private
enum Inner {
    #[cfg(feature = "web-sys-0-3")]
    V0_3(web_sys_0_3::JsValue),
    #[cfg(feature = "web-sys-0-4")]
    V0_4(web_sys_0_4::JsValue),
}

impl Wbg02CanvasWindowHandle {
    #[cfg(feature = "web-sys-0-3")]
    pub fn new_0_3(js_value: web_sys_0_3::JsValue) -> Self {
        Self(Inner::V0_3(js_value))
    }

    #[cfg(feature = "web-sys-0-4")]
    pub fn new_0_4(js_value: web_sys_0_4::JsValue) -> Self {
        Self(Inner::V0_4(js_value))
    }

    #[cfg(feature = "web-sys-0-3")]
    pub fn value_0_3(&self) -> Option<js_value: web_sys_0_3::JsValue> {
        match self.inner {
            Inner::V0_3(value) => Some(value),
            _ => None
        }
    }

    #[cfg(feature = "web-sys-0-4")]
    pub fn value_0_4(&self) -> Option<js_value: web_sys_0_4::JsValue> {
        match self.inner {
            Inner::V0_4(value) => Some(value),
            _ => None
        }
    }
}
```

notgull:
If we were to go this direction, I would prefer to use `JsValue` instead of `web-sys` types directly in order to avoid potential version churn.

I've just realized that the additional downside of including types like that would be that it would make it impossible to release `raw-window-handle` version 1.0, and therefore other 1.0 releases of other crates (like, say, `glutin`). Unless `wasm-bindgen` is planning on bumping from v0.2 to v1.0 anytime soon it would essentially trap the entire Rust GUI ecosystem into an unstable version.

I'll still re-iterate my past points, as even with the proposed model it still becomes increasingly more complicated to construct handles with little added benefit (as the downstream consumer is expected to be using `unsafe` anyways, I would hope that they closely read the documentation).

Lokathor:
Getting this to 1.0 is absolutely a goal. However, we don't need to get the whole crate to 1.0 all at once. Even within rust itself it's common for some parts to stabilize before others (eg: inline asm).

We *could* gate the entire browser handle system behind optional dependencies/features and document it as still experimental even while the rest of the crate for desktop/mobile is moved up to 1.0. If necessary the feature can be called something more generic like `browser` if we want to avoid favoring a particular crate.

notgull:
I would like to avoid using a feature in this case. If we hit 1.0, then `wasm-bindgen` or whatever hits 1.0, we'll have a hanging feature that does nothing that we can't get rid of. In this case I would prefer a `cfg` directive, like `--cfg rwh_semver_exempt_web_handles`.

grovesNL:
> Unless `wasm-bindgen` is planning on bumping from v0.2 to v1.0 anytime soon it would essentially trap the entire Rust GUI ecosystem into an unstable version.

For what it's worth, this is already the case in most crates that supporting this target, e.g. [winit](https://github.com/rust-windowing/winit/blob/189a0080a6df9169b549c405e73328d74e83fcbb/src/platform/web.rs#L50), [wgpu](https://github.com/gfx-rs/wgpu/blob/052bd17d41c227628ba2a6b9c56d8f8af299fb8f/wgpu/src/backend/web.rs#L618), [glow](https://github.com/grovesNL/glow/blob/3c06c30ff370452397f49f840013d86a339b3709/src/web_sys.rs#L1152), [egui's eframe](https://github.com/emilk/egui/blob/bdc8795b0476c25faab927fc3c731f2d79f2098f/crates/eframe/src/web/mod.rs#L99), [leptos](https://github.com/leptos-rs/leptos/blob/4001561987f97752195b673beca19b9373629a22/leptos_dom/src/helpers.rs#L9), etc. all expose `web_sys` types in their public API for the same reasons I've tried to outline - it's much more convenient to work with these typed handles than requiring them to go through the wasm ABI or similar, and the dependency will already be in-tree for this target.

Lokathor:
I don't think a special cfg instead of a cargo feature is very user friendly. Users expect to be able to look at the features list and have it really be the list without secret extra options.

As to having a feature that eventually does nothing: It's really not a big deal. If we really wanted we could even remove the feature (since we're not promising the feature is part of our stable API in the first place) if it was really so offense to the sensibilities, but it's the best user experience to just let an opt-in feature "do nothing" once it's eventually part of the default setup.

notgull:
Yeah, now that you put it that way. I'll rewrite this PR to use `wasm-bindgen` then, if everyone else is fine with it.

notgull:
Ah wait, `JsValue` is not `Copy`. That kind of throws a wrench in the entire thing, as we need `RawWindowHandle` to be `Copy`.

So, I think the current implementation of this PR is still the best way of doing it.

grovesNL:
Ah that's too bad. Opened #138 to see if it's worthwhile to consider lifting that.

notgull:
I've now replaced the wasm-bindgen handles such that they use actual wasm-bindgen objects. This also solves the `Copy` issue mentioned in #142

notgull:
cc @grovesNL

Cargo.toml
```diff
@@ -15,6 +15,21 @@ rust-version = "1.64"
...
 std = ["alloc"]
 
+# Unstable feature to allow the use of `wasm-bindgen` in the `wasm` target.
+# 
+# This is radioactive semver-wise. Do not rely on this in stable code. Any code enabled by this
+# feature can be changed or removed in a breaking way at any time. In fact, this feature could
+# be removed entirely tomorrow.
+# 
+# The point being: rely on this at your own peril.
+unstable_web_handles = ["wasm-bindgen", "std"]
```
madsmtm, line 25: You're already naming the types for `wasm-bindgen` `v0.2`; I think the feature name should include the version as well?

src/web.rs
```diff
@@ -48,3 +50,152 @@ impl WebWindowHandle {
...
+/// uninhabited.
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+/// [`OffscreenCanvas`]: https://docs.rs/web-sys/latest/web_sys/struct.OffscreenCanvas.html
+#[derive(Clone, PartialEq)]
+pub struct Wbg02Object {
+    /// For when `unstable_web_handles` is enabled, this is the index of the object in the
+    /// `wasm-bindgen` table.
+    #[cfg(all(target_family = "wasm", feature = "unstable_web_handles"))]
+    idx: wasm_bindgen::JsValue,
```
madsmtm, line 138: So, I tried to read the above discussion, but I still don't understand why we don't store the value in a way that it can be `Copy`?

Taking my purist hat on for a second, this would be really nice because then `WindowHandle<'a>` really _would_ be a borrowed handle.

<details>
<summary>I can think of at least three ways to solve that</summary>

### Just use a pointer

Just use a pointer to `JsValue`. E.g. we'd store `NonNull<JsValue>`, and let the implementer ensure that the pointer is valid for the lifetime of the implementer.

### Transmute the bytes

Unsafely using something like `val: [u8; mem::size_of::<JsValue>()]`.
On initialization we'd take a `&JsValue` and copy the bytes, and to pass back a reference we'd cast a pointer to the bytes to a `*const JsValue`, and then convert that to a reference.

Though of course this assumes a few things about `JsValue`, like that it doesn't have interior mutability, but still less than if we store a `u32`.

### Don't have raw handle on WASM

We could have platform-specific raw and borrowed handles, e.g. `AppkitWindowHandle<'a> { raw: RawAppkitWindowHandle, p: PhantomData<&'a ()> }`. This would allow the web backend to expose `WasmWindowHandle<'a>(&'a JsValue)`, but not the raw one, if people don't like pointers.
</details>

I agree that even nicer would be to have something that owns the window handles, and `Drop`s them when necessary - if we took a page out of `std::os::fd`'s book, something like an `OwnedWindowHandle` that itself implements `HasWindowHandle`. This would be useful on _all_ platforms!

But the problem with that is, again, dependencies, which is something that `raw-window-handle` tries hard to avoid.

madsmtm (changes requested):

src/web.rs
grovesNL: Could you clarify the benefits of making it `Copy` internally? What’s the use case for reusing a window handle multiple times where cloning wouldn’t be reasonable?

> let the implementer ensure that the pointer is valid for the lifetime of the implementer

As described above, this would break common use cases where people temporarily get a JsValue for a canvas somewhere then drop it after handing it off to the graphics API. They could work around this by leaking the JsValue then using a pointer. This would be a significant  downgrade from how people already manage this today, so I don't expect people to go through raw window handle if that's the case.

> Transmute the bytes

Conceptually JsValue is Arc-like, so it has all the same downsides of pretending Arc is Copy and the same problems as the WasmAbi functions as described above. e.g., you’re effectively creating as many weak references as you want but not tracking this anywhere.

notgull: One of the main sticking points above is that making the WASM handle `Copy` leads to footguns, so I'd rather avoid this if possible.

madsmtm: I think having `Copy` is more from a purist perspective, not _that_ important.

But the great value of the pointer solution is that once we have a pointer `NonNull<JsValue>`, we can just do `NonNull<c_void>`, and then avoid the dependency altogether (ofc we'd still have to version it in the type, but we need not have the actual dependency, which is great for compile times and such).

> As described above, this would break common use cases where people temporarily get a JsValue for a canvas somewhere then drop it after handing it off to the graphics API. They could work around this by leaking the JsValue then using a pointer. This would be a significant downgrade from how people already manage this today, so I don't expect people to go through raw window handle if that's the case.

I'm not that familiar with WASM, but it seems like rendering libraries would have to retain the `JsValue` themselves if they are to use it past the lifetime of the `WindowHandle<'a>` they've been given, right?

E.g. let's assume a rendering library function `fn render(&self, handle: WindowHandle<'a>) -> ...` that renders into a provided canvas. They have basically three options:
- Render right away.
- Store the `WindowHandle<'a>`, in which case they either have to carry the lifetime `'a` along, preventing the parent `JsValue` used to construct the window handle from being dropped until they do.
- Clone the `JsValue`, and manage the lifetime themselves.

In all cases a pointer to `JsValue` (or to `HtmlCanvasElement`, or whatever) seem perfectly fine to me? Or am I missing something?

madsmtm:
Another argument for keeping `raw-window-handle` completely dependency-free: It can more easily be integrated into other projects, e.g. maybe `web-sys` should be the one depending on `raw-window-handle`?

Just a thought, I've at least thought about adding support for it in `icrate`, but have held off because `raw-window-handle` doesn't _guarantee_ it won't have any dependencies (which is very important for low-level libraries to avoid a cyclic dependency in the future).

daxpedda:
> [..] maybe `web-sys` should be the one depending on `raw-window-handle`?

It is unlikely to happen, `web-sys` is supposed to be bindings only, even Serde support was removed/deprecated. It's a `*-sys` crate after all.

src/web.rs
grovesNL: > (ofc we'd still have to version it in the type, but we need not have the actual dependency, which is great for compile times and such).

The dependency will almost always (I'd guess >99% of the time) already be in-tree on either or both sides (producer and consumer of the handle) for this target, so I'm not sure about the trade-off of avoiding a dependency that's likely to be in-tree by removing type safety and adding pointer casts.

> In all cases a pointer to JsValue (or to HtmlCanvasElement, or whatever) seem perfectly fine to me? Or am I missing something?

That could work as long as the consumer is able to dereference and clone the `JsValue`, it's just extra pointer casts for producers and consumers to get wrong.

notgull: > But the great value of the pointer solution is that once we have a pointer `NonNull<JsValue>`, we can just do `NonNull<c_void>`, and then avoid the dependency altogether (ofc we'd still have to version it in the type, but we need not have the actual dependency, which is great for compile times and such).

I'm against having opaque pointers to Rust types. It's possible for two versions of `wasm-bindgen` to be compiled into an executable, and unless the type is explicitly defined like it is here it's impossible to tell which version it's pointing to.

notgull:
> Just a thought, I've at least thought about adding support for it in `icrate`, but have held off because `raw-window-handle` doesn't _guarantee_ it won't have any dependencies (which is very important for low-level libraries to avoid a cyclic dependency in the future).

The dependency is only present in web targets; I am committed to maintaining `raw-window-handle`'s dependency-free status for desktop and mobile platforms. So it should still be fine to add to `icrate`.

Lokathor:
> It's a *-sys crate after all.

Even a sys crate usually declares types, and thus can implement appropriate traits on those types. Not that it has to depend on us and do that, but it would be reasonable

daxpedda:
> Even a sys crate usually declares types, and thus can implement appropriate traits on those types. Not that it has to depend on us and do that, but it would be reasonable

Wasn't a question of "can", but of "should". My impression so far is that `*-sys` crates tend to have as few dependencies as possible and let others depend on them instead of the other way around.

This isn't a rule somewhere, so I don't think it's totally unreasonable.
Would anybody be interested to make a PR (when we get there) to `web-sys` so that I can review it and discuss it with the others in Rustwasm? Myself I'm not familiar at all with `raw-window-handle`.

src/web.rs
daxpedda: You can't have two versions of `wasm-bindgen` compiled into the same executable, other then the fact that it wouldn't work anyway, the CLI tool prevents it.

madsmtm: > The dependency will almost always (I'd guess >99% of the time) already be in-tree on either or both sides (producer and consumer of the handle) for this target, so I'm not sure about the trade-off of avoiding a dependency that's likely to be in-tree by removing type safety and adding pointer casts.

> it's just extra pointer casts for producers and consumers to get wrong.

Very true! Indeed it _is_ a trade-off between "ease of use"/"safety" and "no dependencies"/"consistency". That trade-off was made with the other platforms, where the problem of dependencies are arguably worse, and where the ABI story is better.

My opinion is still that the better choice is "no dependencies", partly because of consistency, partly for the other reasons described, but I can see why we disagree.

> and unless the type is explicitly defined like it is here it's impossible to tell which version it's pointing to.

Hmm, I don't see how `Wbg02CanvasWindowHandle(NonNull<c_void>)` or something like that could possibly refer to multiple versions of `wasm-bindgen`? Though I get that people could accidentally forget to update the name when they update the dependency, and there'd be no safeguards against that.

notgull: > My opinion is still that the better choice is "no dependencies", partly because of consistency, partly for the other reasons described, but I can see why we disagree.

As much as the idea of having zero dependencies appeals to me, I'm leaning towards the current option. If only because the footguns introduced otherwise would be a deal breaker for `wgpu` (I assume that @grovesNL is speaking for them).

> Hmm, I don't see how `Wbg02CanvasWindowHandle(NonNull<c_void>)` or something like that could possibly refer to multiple versions of `wasm-bindgen`? Though I get that people could accidentally forget to update the name when they update the dependency, and there'd be no safeguards against that.

Turns out that it can't (thanks Daxpedda), but it would be an issue for other pure Rust handles.

madsmtm: Okay, how about this then:
```rust
pub enum RawWindowHandle {
    WebOffscreenCanvas(WebOffscreenCanvasWindowHandle),
    WebCanvas(WebCanvasWindowHandle),
    // ...
}

pub struct WebOffscreenCanvasWindowHandle {
    /// The `JsValue` of an `OffscreenCanvas`.
    pub value: NonNull<c_void>,
}

#[cfg(feature = "wasm-bindgen-0-2")]
impl WebOffscreenCanvasWindowHandle {
    pub fn from_0_2(canvas: &wasm_bindgen::JsValue) -> Self {
        Self { value: NonNull::from(canvas).cast::<c_void>() }
    }

    // Safe if the handle is gotten through `WindowHandle<'a>`, which ensures validity of the pointers and such
    pub unsafe fn get_0_2(&self) -> &wasm_bindgen::JsValue {
        unsafe { self.value.cast::<wasm_bindgen::JsValue>().as_ref() }
    }
}

// Or for even more type-safety
#[cfg(feature = "web-sys-0-3")]
impl WebOffscreenCanvasWindowHandle {
    pub fn from_web_sys_0_3(canvas: &web_sys::OffscreenCanvas) -> Self {
        let value: &wasm_bindgen::JsValue = &canvas; // Deref
        Self::from_0_2(value)
    }

    // Safe if the handle is gotten through `WindowHandle<'a>`, which ensures validity of the pointers and such.
    pub unsafe fn get_web_sys_0_3(&self) -> &web_sys::OffscreenCanvas {
        OffscreenCanvas::unchecked_from_js_ref(self.get_0_2())
    }
}

// `HtmlCanvasElement`
pub struct WebCanvasWindowHandle { pub value: NonNull<c_void> }

// Similar impls for `WebCanvasWindowHandle`
```

Since `wasm-bindgen` is already ensured to be unique, there's no need for the versioning scheme on the name.

This would give us `Copy` back, we wouldn't have variants that directly depend on `cfg`s, and it would allow `raw-window-handle` to more easily handle upgrades to the dependency (and would allow windowing libraries and graphics libraries to do the casts themselves, leading to faster compile-times by unblocking `raw-window-handle` in the compilation graph, since it can avoid depending on `web-sys` directly).

notgull: That sounds fine to me. @grovesNL Does this work for `wgpu`?

grovesNL: That would work ok for wgpu and I definitely don't want to block progress here.

I personally think this is the wrong tradeoff (e.g. very real risk of use-after-free as we've seen with people trying ash for the first time) and we shouldn't be forcing people to deal with pointer casting type safe handles to c_void just to always recover the same type on the other side, especially because it's a downgrade from what people are already doing (passing type safe handles directly to graphics APIs). The original goal of raw-window-handle was to make interop easier between crates, but this makes it harder to get right.

notgull: Agreeing with the above, in addition to the fact that casting Rust types to and from a C void pointer feels very unclean to me.

madsmtm: I think we should go for this design then for now - I am definitely open for discussing further, though we should probably do so in a separate issue (esp. after we've seen how the design of `WindowHandle<'_>` actually works out in practice).

@notgull do you want to take a stab at it in this PR?

notgull: I've changed it to this for now. I'll open an issue once this PR is merged about this.

Cargo.toml
```diff
@@ -15,6 +15,21 @@ rust-version = "1.64"
 alloc = []
 std = ["alloc"]
 
+# Unstable feature to allow the use of `wasm-bindgen` in the `wasm` target.
+# 
+# This is radioactive semver-wise. Do not rely on this in stable code. Any code enabled by this
+# feature can be changed or removed in a breaking way at any time. In fact, this feature could
+# be removed entirely tomorrow.
+# 
+# The point being: rely on this at your own peril.
```
madsmtm, line 24: I know that I've been vocal about it, but honestly, if we have the dependency anyhow, we might as well make it stable, no? I mean, `raw-window-handle` is not yet stable, so we can always deprecate an remove it later if we feel the need to.

src/web.rs
```diff
@@ -48,3 +51,159 @@ impl WebWindowHandle {
...
+    /// of [`HtmlCanvasElement`]. The pointer is a direct reference to a [`JsValue`].
+    ///
+    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
+    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    /// [`JsValue`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html
+    pub obj: NonNull<c_void>,
+}
+
+impl Wbg02CanvasWindowHandle {
+    /// Create a new handle to an [`HtmlCanvasElement`].
```
madsmtm, line 83: Nit: Make it clear that we're dealing with a pointer

```suggestion
    /// Create a new handle to a pointer to a [`HtmlCanvasElement`].
```

src/web.rs
```diff
@@ -48,3 +51,159 @@ impl WebWindowHandle {
...
+/// ```
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Wbg02CanvasWindowHandle {
+    /// The object representing the [`HtmlCanvasElement`].
+    ///
+    /// It is implied that this object is registered in the [`wasm-bindgen`] table and is an instance
+    /// of [`HtmlCanvasElement`]. The pointer is a direct reference to a [`JsValue`].
```
madsmtm, line 74: Nit: wording ("it is implied" sounds odd).

```suggestion
    /// A pointer to the [`JsValue`] of a [`HtmlCanvasElement`].
    ///
    /// The object must be registered in the [`wasm-bindgen`] table.
```

(Unsure what you meant my "registered", so I kept it)

src/web.rs
```diff
@@ -48,3 +51,159 @@ impl WebWindowHandle {
...
+/// # fn get_canvas() -> NonNull<c_void> { unimplemented!() }
+/// let obj: NonNull<c_void> = get_canvas();
+/// let mut window_handle = Wbg02CanvasWindowHandle::new(obj);
+/// /* set fields */
+/// ```
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct Wbg02CanvasWindowHandle {
```
madsmtm, line 70: Name doesn't have to include version info, since there can only be one instance of `wasm-bindgen` (though we should definitely document that assumption)

src/web.rs
```diff
@@ -48,3 +51,159 @@ impl WebWindowHandle {
...
+    ///
+    /// The [`JsValue`] must refer to an [`HtmlCanvasElement`], and the lifetime must be longer than
+    /// the `Wbg02CanvasWindowHandle` lives for.
+    ///
+    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(target_family = "wasm", feature = "unstable_web_handles_wbg_02")))
+    )]
+    pub unsafe fn from_wasm_bindgen_0_2(js_value: &wasm_bindgen::JsValue) -> Self {
```
madsmtm, line 109: This function doesn't need to be `unsafe`, the safety guarantees are upheld in `WindowHandle::new`

src/web.rs
```diff
@@ -48,3 +51,159 @@ impl WebWindowHandle {
...
+    ///
+    /// # Safety
+    ///
+    /// The [`JsValue`] must refer to an [`HtmlCanvasElement`], and the lifetime must be longer than
+    /// the `Wbg02CanvasWindowHandle` lives for.
+    ///
+    /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+    #[cfg_attr(
+        docsrs,
+        doc(cfg(all(target_family = "wasm", feature = "unstable_web_handles_wbg_02")))
```
madsmtm, line 107: Nit: I suspect placing this on the `impl` instead would have the same effect in the docs?

madsmtm (changes requested):
Nice, thanks @notgull!

Cargo.toml
```diff
@@ -15,6 +15,21 @@ rust-version = "1.64"
...
+unstable_web_handles_wbg_02 = ["wasm-bindgen", "std"]
+
+[target.'cfg(target_family = "wasm")'.dependencies.wasm-bindgen]
+version = "0.2.87"
+default-features = false
+features = ["std"]
```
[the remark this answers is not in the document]
notgull: `wasm-bindgen` fails to compile without it.

Cargo.toml
notgull: If we make it not unstable, then if we upgrade to version 1.0 we can't do the semver trick in 0.6 so backport these changes.

src/web.rs
notgull: If `wasm-bindgen` 1.0 is released, then it would be a breaking change to change it. So for semver reasons it should be versioned.

Cargo.toml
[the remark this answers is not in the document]
madsmtm: Ah, fair enough. Why do we need to enable `"std"` for ourselves then?

Cargo.toml
madsmtm: Well, nowadays we can't do the semver trick anyhow...

But anyhow, maybe that's just how it'll have to be? Like, then we release `v1.0` of `raw-window-handle`, and just forever expose a (likely deprecated when future `wasm-bindgen` versions come out) way to access `wasm-bindgen` `v0.2` traits. Annoying, sure, but better than the code of users of this feature (of which there will be many) failing to compile suddenly after a `cargo update`, no?

Cargo.toml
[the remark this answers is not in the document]
notgull: I think it's so `wasm-bindgen` can disable `std` in the future once they stop depending on it.

src/web.rs
madsmtm: I don't think it would, since if there can only be one version of `wasm-bindgen` in a given binary.

Like, the `NonNull<c_void>` just points to a `wasm_bindgen::JsValue`, whether that is currently `v0.2`, `v0.3` or `v1.0` is irrelevant - it's ensured to always be the same between the windowing and graphics crate you're using, right?

src/borrowed.rs
```diff
@@ -84,7 +84,7 @@ impl<H: HasDisplayHandle + ?Sized> HasDisplayHandle for alloc::sync::Arc<H> {
 ///
 /// Get the underlying raw display handle with the [`HasRawDisplayHandle`] trait.
 #[repr(transparent)]
-#[derive(PartialEq, Eq, Hash, Copy, Clone)]
```
madsmtm, line 87: Rather do the opposite; re-add `Copy` to `WindowHandle<'_>`, now that we can again.

src/web.rs
notgull: Yeah that makes sense, do you mind making the change? I'm at a football game right now (friends think I am scrolling Twitter) so I won't have laptop access for a few hours

Cargo.toml
notgull: I see your point. Then, just make it a stable feature.

src/web.rs
```diff
@@ -48,3 +51,145 @@ impl WebWindowHandle {
...
+    ///
+    /// ## Example
+    /// ```no_run
+    /// # use raw_window_handle::Wbg02CanvasWindowHandle;
+    /// # use core::{ffi::c_void, ptr::NonNull};
+    /// # fn get_canvas() -> NonNull<c_void> { unimplemented!() }
+    /// let obj: NonNull<c_void> = get_canvas();
+    /// let mut window_handle = Wbg02CanvasWindowHandle::new(obj);
+    /// /* set fields */
+    /// ```
```
madsmtm, line 82: Nit: No fields to set later, make it actually explain a little bit about what to do, and let CI run it.

```suggestion
    /// # Example
    ///
    /// ```
    /// # use core::ffi::c_void;
    /// # use core::ptr::NonNull;
    /// # use raw_window_handle::Wbg02CanvasWindowHandle;
    /// # type HtmlCanvasElement = ();
    /// # type JsValue = ();
    /// let canvas: &HtmlCanvasElement;
    /// # canvas = &();
    /// let value: &JsValue = &canvas; // Deref to `JsValue`
    /// let obj: NonNull<c_void> = NonNull::from(value).cast();
    /// let mut handle = Wbg02CanvasWindowHandle::new(obj);
    /// ```
```

src/web.rs
```diff
@@ -48,3 +51,145 @@ impl WebWindowHandle {
...
+    doc(cfg(all(target_family = "wasm", feature = "unstable-wasm-bindgen-0-2")))
+)]
+/// These implementations are only available when `unstable-wasm-bindgen-0-2` is enabled.
+impl Wbg02CanvasWindowHandle {
+    /// Create a new `Wbg02CanvasWindowHandle` from a [`wasm-bindgen`] object.
+    ///
+    /// This function is unstable. Its signature may be changed or even removed outright without a
+    /// breaking version change.
+    ///
+    /// # Safety
```
madsmtm, line 100: Nit: Not a safety comment, though still keep the comment

src/web.rs
madsmtm: Gladly!

Cargo.toml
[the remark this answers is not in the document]
madsmtm: I meant, why are we enabling `raw-window-handle`'s own `"std"` feature when the `wasm-bindgen` feature is enabled?

src/web.rs
```diff
@@ -48,3 +51,131 @@ impl WebWindowHandle {
...
+    ///
+    /// ```
+    /// # use core::ffi::c_void;
+    /// # use core::ptr::NonNull;
+    /// # use raw_window_handle::WebOffscreenCanvasWindowHandle;
+    /// # type OffscreenCanvas = ();
+    /// # type JsValue = ();
+    /// let canvas: &OffscreenCanvas;
+    /// # canvas = &();
+    /// let value: &JsValue = &canvas; // Deref to `JsValue`
```
madsmtm, line 147: @daxpedda: Is the ABI of `OffscreenCanvas` and `JsValue` guaranteed to be the same? E.g. is it sound to transmute `&OffscreenCanvas -> &JsValue`?

src/web.rs
```diff
@@ -48,3 +51,131 @@ impl WebWindowHandle {
...
+/// Raw window handle for a Web canvas registered via [`wasm-bindgen`].
+///
+/// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct WebCanvasWindowHandle {
+    /// A pointer to the [`JsValue`] of an [`HtmlCanvasElement`].
+    ///
+    /// [`JsValue`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html
+    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
```
madsmtm, line 64: @daxpedda: Do you have a reference for this? How guaranteed is it that there will never be two `wasm-bindgen`?

src/web.rs
```diff
@@ -48,3 +51,147 @@ impl WebWindowHandle {
...
+    /// directly.
+    ///
+    /// [`JsValue`]: https://docs.rs/wasm-bindgen/latest/wasm_bindgen/struct.JsValue.html
+    /// [`HtmlCanvasElement`]: https://docs.rs/web-sys/latest/web_sys/struct.HtmlCanvasElement.html
+    //
+    // SAFETY: Not using `JsValue` is sound because `wasm-bindgen` guarantees
+    // that there's only one version of itself in any given binary, and hence
+    // we can't have a type-confusion where e.g. one library used `JsValue`
+    // from `v0.2` of `wasm-bindgen`, and another used `JsValue` from `v1.0`;
+    // the binary will simply fail to compile!
```
madsmtm, line 73: @daxpedda: Do you have a reference for this? How guaranteed is it that there will never be two `wasm-bindgen`?

src/web.rs
daxpedda: Looking at this code, the `OffscreenCanvas` is first dereferenced to `JsValue`, that is then converted to a pointer. If you convert that pointer back to a `JsValue`, that works. I wouldn't convert the `JsValue` pointer to a `OffscreenCanvas` value, same rules as Rust applies here.

Unless `OffscreenCanvas` is a `#[repr(transparent)]` wrapper around `JsValue`, it won't work. Which currently it does, but it doesn't say anywhere that this is stable stuff a user can rely on. I honestly have no clue if Rust API guidelines says that this stuff should be stable in libraries or not.

src/web.rs
daxpedda: It is not 100% guaranteed.

But generally speaking the way `wasm-bindgen` works prevents this, because it encodes "stuff" into the Wasm binary that needs to be interpreted by the CLI tool, so multiple versions will always be incompatible unless `wasm-bindgen` becomes a completely different beast.

That said, it was discussed in the past to move `wasm-bindgen` to the component model (which isn't stable yet) and drop the CLI completely at some point. This would probably allow multiple versions of `wasm-bindgen` to co-exist in the same binary, but old versions will still be unusable. But considering that `wasm-bindgen` is not actively developed anymore and there are now things like `wit-bindgen`, I doubt this will ever happen.

```rust
    // the binary will simply fail to compile!
```

I believe it should compile, but you won't be able to use it. Without the CLI tool it won't work on any runtime and the CLI tool will fail if you have multiple versions.

src/web.rs
madsmtm: Yeah, I kinda figured the `#[repr(transparent)]` wasn't a stable promise (though the attribute may show up in `rustdoc`, it's not a stable promise unless the docs say it is).

Cargo.toml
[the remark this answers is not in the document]
notgull: Huh, I remember it not compiling, but apparently it does? Odd.

notgull (commented):
+1 to all of your changes @madsmtm

I'd be fine with merging now and bikeshedding later

src/web.rs
daxpedda: FWIW: it shows up in the [documentation](https://docs.rs/web-sys/0.3.64/web_sys/struct.OffscreenCanvas.html).

PR #136, opened 2023-07-28
Change how handles are constructed, so that we can ensure non-null handles

madsmtm:
A _lot_ of consumers of `raw-window-handle` assume that the pointers are non-null, which makes those libraries unsound. Since the traits are fallible nowadays, we should try to use the type-system to ensure that the handles are always completely valid, instead of expecting user to check it themselves.

This could be done by using `Option<NonNull<c_void>>`, but that still needs the user to unwrap what _should_ always be set anyhow, so I opted for renaming the `empty` methods to `new`, and taking parameters to these where it makes sense. Note that the structs are still left open to non-breaking extensions, as they are still marked `#[non_exhaustive]` - with the caveat that those extensions must be optional / have a sensible default.

See https://github.com/rust-windowing/raw-window-handle/pull/33 and https://github.com/rust-windowing/raw-window-handle/pull/55 for a bit of previous discussion on this.

### DIFF
--- a/src/haiku.rs
+++ b/src/haiku.rs
@@ -1,46 +1,57 @@
 use core::ffi::c_void;
-use core::ptr;
+use core::ptr::NonNull;
 
 /// Raw display handle for Haiku.
-///
-/// ## Construction
-/// ```
-/// # use raw_window_handle::HaikuDisplayHandle;
-/// let mut display_handle = HaikuDisplayHandle::empty();
-/// /* set fields */
-/// ```
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct HaikuDisplayHandle;
+pub struct HaikuDisplayHandle {}
 
 impl HaikuDisplayHandle {
-    pub fn empty() -> Self {
+    /// Create a new empty display handle.
+    ///
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use raw_window_handle::HaikuDisplayHandle;
+    /// let handle = HaikuDisplayHandle::new();
+    /// ```
+    pub fn new() -> Self {
         Self {}
     }
 }
 
 /// Raw window handle for Haiku.
-///
-/// ## Construction
-/// ```
-/// # use raw_window_handle::HaikuWindowHandle;
-/// let mut window_handle = HaikuWindowHandle::empty();
-/// /* set fields */
-/// ```
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct HaikuWindowHandle {
     /// A pointer to a BWindow object
-    pub b_window: *mut c_void,
+    pub b_window: NonNull<c_void>,
     /// A pointer to a BDirectWindow object that might be null
-    pub b_direct_window: *mut c_void,
+    pub b_direct_window: Option<NonNull<c_void>>,
 }
 
 impl HaikuWindowHandle {
-    pub fn empty() -> Self {
+    /// Create a new handle to a window.
+    ///
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use core::ptr::NonNull;
+    /// # use raw_window_handle::HaikuWindowHandle;
+    /// # type BWindow = ();
+    /// #
+    /// let b_window: NonNull<BWindow>;
+    /// # b_window = NonNull::from(&());
+    /// let mut handle = HaikuWindowHandle::new(b_window.cast());
+    /// // Optionally set `b_direct_window`.
+    /// handle.b_direct_window = None;
+    /// ```
+    pub fn new(b_window: NonNull<c_void>) -> Self {
         Self {
-            b_window: ptr::null_mut(),
-            b_direct_window: ptr::null_mut(),
+            b_window,
+            b_direct_window: None,
         }
     }
 }
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -7,7 +7,10 @@ use core::ptr::NonNull;
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct XlibDisplayHandle {
     /// A pointer to an Xlib `Display`.
-    pub display: NonNull<c_void>,
+    ///
+    /// It is strongly recommended to set this value, however it may be set to
+    /// `None` to request the default display when using EGL.
+    pub display: Option<NonNull<c_void>>,
 
     /// An X11 screen to use with this display handle.
     ///
@@ -32,9 +35,9 @@ impl XlibDisplayHandle {
     /// let screen;
     /// # display = NonNull::from(&()).cast();
     /// # screen = 0;
-    /// let handle = XlibDisplayHandle::new(display, screen);
+    /// let handle = XlibDisplayHandle::new(Some(display), screen);
     /// ```
-    pub fn new(display: NonNull<c_void>, screen: c_int) -> Self {
+    pub fn new(display: Option<NonNull<c_void>>, screen: c_int) -> Self {
         Self { display, screen }
     }
 }
@@ -78,7 +81,10 @@ impl XlibWindowHandle {
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct XcbDisplayHandle {
     /// A pointer to an X server `xcb_connection_t`.
-    pub connection: NonNull<c_void>,
+    ///
+    /// It is strongly recommended to set this value, however it may be set to
+    /// `None` to request the default display when using EGL.
+    pub connection: Option<NonNull<c_void>>,
 
     /// An X11 screen to use with this display handle.
     ///
@@ -103,9 +109,9 @@ impl XcbDisplayHandle {
     /// let screen;
     /// # connection = NonNull::from(&()).cast();
     /// # screen = 0;
-    /// let handle = XcbDisplayHandle::new(connection, screen);
+    /// let handle = XcbDisplayHandle::new(Some(connection), screen);
     /// ```
-    pub fn new(connection: NonNull<c_void>, screen: c_int) -> Self {
+    pub fn new(connection: Option<NonNull<c_void>>, screen: c_int) -> Self {
         Self { connection, screen }
     }
 }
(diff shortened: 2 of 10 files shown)


kchibisov:
I understand the motivation behind that and I agree that it's sort of useful to do so. Certain data is mandatory for a handle to be valid and useful in the first place.

I'm just not sure how fragile it could be in the wild. Some pointers could be nullable, like display on X11, because users could try to get some sort of default.

madsmtm:
Yeah, I definitely need everybody's guidance on where a null pointer does or does not make sense (e.g. a null pointer NSView is definitely nonsense, there's nothing you can feasibly fall back to, but for others it may).

src/haiku.rs
```diff
@@ -1,46 +1,57 @@
...
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct HaikuWindowHandle {
     /// A pointer to a BWindow object
-    pub b_window: *mut c_void,
+    pub b_window: NonNull<c_void>,
     /// A pointer to a BDirectWindow object that might be null
-    pub b_direct_window: *mut c_void,
+    pub b_direct_window: Option<NonNull<c_void>>,
 }
```
notgull, line 32: Do we have anyone experienced with Haiku available? I'm not sure whether or not this is a good idea.

src/windows.rs
```diff
@@ -1,71 +1,84 @@
...
-/// /* set fields */
-/// ```
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Win32WindowHandle {
     /// A Win32 `HWND` handle.
-    pub hwnd: *mut c_void,
-    /// The `HINSTANCE` associated with this type's `HWND`.
-    pub hinstance: *mut c_void,
+    pub hwnd: isize,
```
notgull, line 31: We could use `NonZeroIsize` here.

src/windows.rs
```diff
@@ -1,71 +1,84 @@
...
+    /// The `GWLP_HINSTANCE` associated with this type's `HWND`.
+    pub hinstance: isize,
```
notgull, line 33: Ditto but with `Option<NonZeroIsize>`

src/windows.rs
```diff
@@ -1,71 +1,84 @@
...
+    /// # struct HWND(isize);
+    /// #
+    /// let window: HWND;
+    /// # window = HWND(0);
+    /// # #[cfg(only_for_showcase)]
+    /// let hinstance = unsafe { GetWindowLongPtrW(window, GWLP_HINSTANCE) };
+    /// # let hinstance = 0;
+    /// let mut handle = Win32WindowHandle::new(window.0, hinstance);
+    /// ```
+    pub fn new(hwnd: isize, hinstance: isize) -> Self {
```
notgull, line 53: `hinstance` isn't a necessary field here.

kchibisov:
>By the way, it's generally a bad idea to just "open" a new X11 display if you can't get a current one. wgpu just encountered a bug because of that, though I forget the issue number.

You can create a glx display without a present display, just saying, the same with EGL, you can use EGL_DEFAULT_DISPLAY and it's a valid thing to do on X11 and sort of works(at least that's how it was initially iirc).

I think it's a better to use `Option<NonNull>` and similar instead of raw values or `Option<NonZero>` the null checks are mandatory in the clients anyway, so not sure why we have them implicit here.

@madsmtm you can set `NonNull` for all the wayland stuff.

src/windows.rs
madsmtm: I chose not to, since the `windows` crate [doesn't either](https://microsoft.github.io/windows-docs-rs/doc/windows/Win32/Foundation/struct.HWND.html), but I'm unsure here.

src/haiku.rs
madsmtm: I mostly just followed the docs saying "that might be null"

notgull: Sounds right to me. @hoanga You originally added Haiku support, what do you think?

src/windows.rs
notgull: The `windows` crate is just ensuring maximum FFI compatibility I think, no valid `HWND` is zero. Same goes for `HINSTANCE`.

madsmtm:
> I think it's a better to use `Option<NonNull>` and similar instead of raw values

Agreed, that's also what I've set out to do.

> set `NonNull` for all the wayland stuff.

Done.

> You can create a glx display without a present display

So `XlibDisplayHandle.display` should be `Option<NonNull<c_void>>`? What about `XcbDisplayHandle.connection`, should that also be nullable?

madsmtm:
Tagging @Lokathor, you may remember the previous discussion on this topic better than I

notgull:
> So `XlibDisplayHandle.display` should be `Option<NonNull<c_void>>`? What about `XcbDisplayHandle.connection`, should that also be nullable?

I would personally be against this, for the reasoning above. This API affords itself to be misused.

Lokathor:
I am mildly against this because NonNull is more of a pain in the butt to work with. Using `Option<NonNull<T>>` has worse ergonomics than just `*mut T` much of the time, particularly because actual FFI calls are generally bound as using the raw pointer and so you get `nn.get()` all over the place, because there's no coercion.

However, if everyone else really want to use NonNull I won't block the change.

kchibisov:
But `Option<NonNull>` is FFI friendly though, as in you can pass it as is? The same with other `Option<NonZero>` types.

kchibisov:
The downstream already must validate before using data, but with nonnull they can pass data as is, because it's just a pointer. And they could even pass pointer, but for example in glutin we check every input we have.

Though, I understand why we have `*mut T`, it's just easier to work with in some cases.

Lokathor:
The actual function bindings almost never accept Option<NonNull>

kchibisov:
Yeah, that's fair, and the cast won't work because it's not trivial, so you have only an option to transmute, I guess.

Lokathor:
`opt.map(NonNull::get).unwrap_or(null())` is quite wordy. You can make a fn for it of course, but still.

And I have been advised that, in general, transmute and pointers should not be anywhere near each other, because in some cases that look fine the provenance can become messed up, and it's just much safer to never mix them.

kchibisov:
Yeah, I think raw pointers is fine.

Do you have any opinion wrt plain `NotNull` for an essential data which makes `RawWindowHandle` and having a way to build a handle with the proposed changes?

src/haiku.rs
hoanga: if the general direction is to move away from the current approach, i am not opposed to the change.

as far as i know, no other crates that utilize the haiku path of this library have publicly surfaced so any efforts to update dependencies for this change should be rather minimal.

notgull:
> `opt.map(NonNull::get).unwrap_or(null())` is quite wordy. You can make a fn for it of course, but still.

My opinion is that ensuring greater correctness throughout the `raw-window-handle` ecosystem would take precedence over the ease of writing FFI. In fact, the wordiness here can be easily avoided with an extension trait.

```rust
trait PointerExt<T: ?Sized> {
    fn as_ptr(self) -> *mut T;
}

impl<T: ?Sized> PointerExt<T> for Option<NonNull<T>> {
    fn as_ptr(self) -> *mut T { self.map(NonNull::get).unwrap_or(null()) }
}
```

I am still in favor of the PR in its current form, as it prevents a footgun that is very prevalent throughout the users of `raw-window-handle`.

src/unix.rs
```diff
@@ -79,172 +119,190 @@ pub struct XcbWindowHandle {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use raw_window_handle::DrmDisplayHandle;
+    /// #
+    /// let fd: i32;
+    /// # fd = 0;
+    /// let handle = DrmDisplayHandle::new(fd);
+    /// ```
+    pub fn new(fd: i32) -> Self {
```
kchibisov, line 223: @notgull should we use io safety for that fd? It's a regular fd, if it matters.

notgull: Since putting a `BorrowedFd` in this spot here would require a lifetime as well as a dependency on `libstd`, I would say no.

kchibisov:
Do we want to do anything with XID on X11, zero is not a valid xid if it matters.

notgull:
Yes, I think we should make XID types non-zero as well, at least for XCB, since the Xlib one relies on `std::os::raw`.

src/windows.rs
madsmtm: Moved to https://github.com/rust-windowing/raw-window-handle/pull/137

src/windows.rs
madsmtm: Moved to https://github.com/rust-windowing/raw-window-handle/pull/137

madsmtm:
I've moved the `NonZero*` debate to https://github.com/rust-windowing/raw-window-handle/pull/137, so all that's remaining here is figuring out whether we should wrap `XlibDisplayHandle.display` and `XcbDisplayHandle.connection` in an `Option` or not.

madsmtm:
I went with making the handles nullable, since it's the conservative choice, and then we can discuss that in detail in https://github.com/rust-windowing/raw-window-handle/issues/141 later.

I'll consider this PR ready now, and will merge it in a few days unless someone brings up objections.

src/windows.rs
```diff
@@ -1,71 +1,91 @@
...
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct Win32WindowHandle {
     /// A Win32 `HWND` handle.
-    pub hwnd: *mut c_void,
-    /// The `HINSTANCE` associated with this type's `HWND`.
-    pub hinstance: *mut c_void,
+    pub hwnd: NonZeroIsize,
+    /// The `GWLP_HINSTANCE` associated with this type's `HWND`.
+    pub hinstance: Option<NonZeroIsize>,
```
MarijnS95, line 34: Also for posterity: the type was changed from pointer to `isize` here to match the representation in Windows-rs: https://github.com/microsoft/windows-rs/issues/1643

MarijnS95: FYI, `win32metadata` is changing these back to void pointers it seems:

https://github.com/microsoft/win32metadata/issues/1924
https://github.com/microsoft/win32metadata/commit/b4dfd2f25c0114da8b27d0d390bca5c1555fc55c

See also my local regeneration: https://github.com/microsoft/windows-rs/commit/96b9a27a6293fe08e4a96b64400239594941ff20

CC @kennykerr, @riverar, @mikebattista.

PR #137, opened 2023-07-30
Make integer handles use non-zero types

madsmtm:
In similar spirit as https://github.com/rust-windowing/raw-window-handle/pull/136.

I'm not sure myself that this is such a good idea, mainly because:
- The `c_int/c_long` types are not possible to convert to non-zero types in the same manner (leading to us only using the API in some places).
- Ensuring non-null pointers is _important_, since that causes unsoundness if used incorrectly, while a zero handle isn't that big of a problem, any operations you do on them will just error (just like any other invalid handle).

Other options would be to panic on `0` in the constructor (a weak check against invalid handles, assuming those are exceedingly rare), or we could document that it is a bug in the implementer (as opposed to the consumer) to provide zero handles where unexpected?

### DIFF
--- a/src/unix.rs
+++ b/src/unix.rs
@@ -1,4 +1,5 @@
 use core::ffi::{c_int, c_ulong, c_void};
+use core::num::NonZeroU32;
 use core::ptr::NonNull;
 
 /// Raw display handle for Xlib.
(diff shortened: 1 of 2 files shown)


notgull (approved):
Looks good to me!

notgull:
Oh, my bad, there weren't any comments for a while so I assumed that the debate was settled.

PR #139, opened 2023-08-08
Excise the HasRaw*Handle traits

### LINKED ISSUE #135: Type-system relationship between `HasRaw*Handle` and `Has*Handle` traits
Should there be a type-system relationship between these?

I think it would make sense to make the normal handles subtraits of the raw handles, i.e. `trait HasWindowHandle: HasRawWindowHandle { ... }`?

Or maybe we should do a blanket `impl<T: HasWindowHandle> for HasRawWindowHandle {}`, or vice-versa?

notgull:
After reading #135, I've realized that the HasRawDisplayHandle and HasRawWindowHandle traits have little value in a post-borrowed-handle world. Borrowed handles do everything better, and the raw handle can be extracted from the borrowed handle. Therefore it makes sense to remove these traits.

Closes #135.

### DIFF
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -82,13 +82,6 @@ pub struct DisplayHandle<'a> {
     _marker: PhantomData<&'a *const ()>,
 }
 
-#[allow(deprecated)]
-unsafe impl crate::HasRawDisplayHandle for DisplayHandle<'_> {
-    fn raw_display_handle(&self) -> Result<RawDisplayHandle, HandleError> {
-        Ok(self.raw)
-    }
-}
-
 impl fmt::Debug for DisplayHandle<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_tuple("DisplayHandle").field(&self.raw).finish()
--- a/src/borrowed.rs
+++ b/src/borrowed.rs
@@ -97,6 +97,13 @@ pub struct DisplayHandle<'a> {
     _marker: PhantomData<&'a *const ()>,
 }
 
+#[allow(deprecated)]
+unsafe impl crate::HasRawDisplayHandle for DisplayHandle<'_> {
+    fn raw_display_handle(&self) -> Result<RawDisplayHandle, HandleError> {
+        Ok(self.raw)
+    }
+}
+
 impl fmt::Debug for DisplayHandle<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_tuple("DisplayHandle").field(&self.raw).finish()
@@ -270,11 +277,12 @@ impl<'a> WindowHandle<'a> {
 
     /// Get the underlying raw window handle.
     pub fn as_raw(&self) -> RawWindowHandle {
-        self.raw
+        self.raw.clone()
     }
 }
 
-unsafe impl HasRawWindowHandle for WindowHandle<'_> {
+#[allow(deprecated)]
+unsafe impl crate::HasRawWindowHandle for WindowHandle<'_> {
     fn raw_window_handle(&self) -> Result<RawWindowHandle, HandleError> {
         Ok(self.raw)
     }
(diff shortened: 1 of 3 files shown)


src/borrowed.rs
```diff
@@ -139,8 +163,16 @@ impl<'a> HasDisplayHandle for DisplayHandle<'a> {
 ///
 /// # Safety
```
madsmtm, line 164: It's a bit weird to have a `# Safety` section on a safe trait. I know that the safety guarantees are ensured by the creation method on `WindowHandle`, but maybe it makes sense to document that one more explicitly instead? Not sure!

src/borrowed.rs
```diff
@@ -139,8 +163,16 @@ impl<'a> HasDisplayHandle for DisplayHandle<'a> {
...
+/// Users can safely assume that non-`null`/`0` fields are valid handles, and it is up to the
+/// implementer of this trait to ensure that condition is upheld.
+///
+/// Despite that qualification, implementers should still make a best-effort attempt to fill in all
+/// available fields. If an implementation doesn't, and a downstream user needs the field, it should
+/// try to derive the field from other fields the implementer *does* provide via whatever methods the
+/// platform provides.
+///
+/// The exact handles returned by `raw_window_handle` must remain consistent between multiple calls
+/// to `raw_window_handle` as long as not indicated otherwise by platform specific events.
```
madsmtm, line 175: This one is a little bit dangerous to specify, as people might think they can safely invalidate the window while a borrowed handle is still, well, borrowed, as long as they just emit some kind of event for that, but they cannot!

madsmtm (approved):
This looks perfect honestly.

I could potentially see how an implementer might not want to provide the strong guarantees about validity that `HasWindowHandle`, but `HasRawWindowHandle` _already_ had those guarantees, they were just more implicit, so I think this will be fine!

notgull:
Removing it outright might be a bit too harsh. I've opted to deprecate it for now.

madsmtm:
Maybe you could add a blanket `unsafe impl<T: ?Sized + HasWindowHandle> HasRawWindowHandle for T { ... }`? That would mean less work for crates wanting to implement these traits.

notgull:
> Maybe you could add a blanket `unsafe impl<T: ?Sized + HasWindowHandle> HasRawWindowHandle for T { ... }`? That would mean less work for crates wanting to implement these traits.

Banger idea, I've implemented this

Lokathor:
What would you say people who don't have borrowing lifetimes on their handles do

madsmtm:
> What would you say people who don't have borrowing lifetimes on their handles do

I'm not sure I understand the question, but what would an example of this be? Like, what would be a situation where you, as a "provider" type, would like to say "hey, I have this window handle, its lifetime is not tied to me (`&self`), but I don't know the lifetime of it either"?

Lokathor:
So SDL gives you `*mut SDL_Window` when you make a window.

and because pointers are Copy the "lifetime" of that pointer, as the compiler sees it, is not anything sensible because there's all sorts of spots where you'll accidentally get a copy of the value instead of extending the life of a borrow on the pointer, or some other fiddly thing like that.

So there's certainly *some* span of time that's correct (until you call `SDL_DestroyWindow`), but Rust's borrow checker won't know what the correct time is.

madsmtm:
Makes sense. In this case, _don't_ implement `Has[Raw]WindowHandle` for the pointer, because it _doesn't_ have a window handle for its full lifetime.

Instead, expose an `unsafe` method like this:
```rust
/// # Safety
/// The window must not be destroyed for the lifetime of the handle
pub unsafe fn get_handle<'a>(window: *mut SDL_Window) -> WindowHandle<'a> { ... }
```

Note that `WindowHandle<'_>` implements `HasWindowHandle`, so graphics crates that take `T: HasWindowHandle` will still work. Note also that the lifetime here is unbounded, so the user can choose `'static` if that's needed by their graphics crate.

Lokathor:
Alright, as long as we've thought about it and we have some answer, then we can proceed

notgull:
It seems to me that everyone is fine with this, so I'll merge it.

PR #140, opened 2023-08-11
Remove `Copy` for `RawWindowHandle`/`WindowHandle`

### LINKED ISSUE #138: Does `RawWindowHandle` need to be `Copy`?
It looks like the `Copy` derive was added in #18, I guess because each handle happened to be compatible with it at the time.

However there are some benefits to removing `Copy`, e.g. it makes it possible to include strongly typed web handles that are reference counted (not `Copy`) https://github.com/rust-windowing/raw-window-handle/pull/134#issuecomment-1666306670

grovesNL:
Fixes #138


notgull (approved):
Thanks!

Issue #147, opened 2023-08-31
Add tests for Rust subleties


madsmtm:
There are many small ways to break certain semver guarantees, and it would be nice if we could avoid making these mistakes; so let's add some tests to ensure that!

Add tests for:
- [ ] Presence (or absence) of marker traits on types:
  - `Send`
  - `Sync`
  - `UnwindSafe`
  - `RefUnwindSafe`
  - `Unpin`
- [x] ~Contains interior mutability (e.g. disallowing placing something in a static)~
- [ ] Object safety on the traits we expose.
- [x] Variance of any lifetimes and/or generics we expose.

Possibly also: Integrate [`cargo-semver-checks`](https://github.com/obi1kenobi/cargo-semver-checks) in CI?

Lokathor:
I run cargo-semver-checks on an actual publish but putting it in CI in some way is reasonable as well

kchibisov:
You could do doc tests for all of that like glutin does for example, we test `Send`/`Sync` and `!Send`/`!Sync` there to ensure semantics across the backends.

Issue #85, opened 2021-12-05
Should `RawWindowHandle` implement `Send`?


maroider:
Windows are apparently effectively `!Send` on [Windows and macOS](https://github.com/rust-windowing/raw-window-handle/issues/71#issuecomment-863597284), and the [lifetime issues](https://github.com/rust-windowing/raw-window-handle/issues/71#issuecomment-863584973) that Android's window handles have make marking window handles `Send` perhaps a bit suspect.

I believe we should treat `RawWindowHandle`s somewhat like pointers, in that they can be constructed by safe code, but can't be used directly without `unsafe`. Thus, we should make the handles `Send`, but provide enough information in order for user code to tell if they're using the handle on the correct thread. Additionally, any API creating ready-made `RawWindowHandle`s must be unsafe, as is already the case with `HasRawWindowHandle`. This all assumes we can find a workable solution to #84.

Lokathor:
> I believe we should treat RawWindowHandles somewhat like pointers, in that they can be constructed by safe code, but can't be used directly without unsafe. Thus, we should make the handles Send, but provide enough information in order for user code to tell if they're using the handle on the correct thread.

Except pointers are !Send

And this isn't a super big deal because (unlike Copy) a type can always just force Send onto whatever data it wants.

MaulingMonkey:
I can't speak to macOS, and I don't particularly have a need for `RawWindowHandle` to be `Send`, but I want to chime in on the 
windows side of things conceptually.  I'm going to call a Windows Window a `WND`, even though no such win32 type exists.

* `WND: !Send`, as you say.
* `WND: Sync`, however, per [Raymond Chen / The Old New Thing](https://devblogs.microsoft.com/oldnewthing/20051010-09/?p=33843).
* `HWND` ~ `&WND` / `Option<Weak<WND>>`, where `WND` has a ton of interior mutability:
    * Some Mutex-based (message queues?)
    * Some atomic
    * Some closer to UnsafeCell, caller beware.
* `&T: Send` if `T: Sync`
* ergo it'd be sane for `HWND: Send` given `WND: Sync` (although winapi doesn't implement `Send` for HWND currently.)
* ergo it'd be sane for `RawWindowHandle: Send`, even though `HasRawWindowHandle` likely can't be.

---

[...]

Issue #59, opened 2020-11-14
Should RawWindowHandle be Send?


glowcoil:
It looks like `RawWindowHandle`, and the individual platform handles, are marked `!Send` by default due to the pointers they contain. I couldn't find any documentation one way or another on whether this is explicitly intended to be a part of the safety contract for using a `RawWindowHandle` or not. I assume that there are some use cases where moving a handle between threads is perfectly safe, and others where it is not. Do existing libraries rely on `RawWindowHandle` not implementing `Send` for safety? Or would it be fine for it to impl `Send`?

thomcc:
No, for the same reasons raw pointers aren't. Also, not all platforms support doing windowing operations off of the main (ui) thread.

If you need this and are sure it's sound for your use case/target, you should wrap it in a newtype struct and `unsafe impl Send` for it. I would definitely recommend doing this for a concrete window handle type and not the RawWindowHandle enum, however.

glowcoil:
Sounds reasonable to me, and the newtype approach is what I've done in my own code; just wanted to hear some opinions since it isn't mentioned in the docs.

PR #152, opened 2023-10-13
Make handles Send + Sync

madsmtm:
In `winit` we're currently discussing this, both in https://github.com/rust-windowing/winit/pull/3136 and [on Matrix](https://matrix.to/#/!DGpLzJTRzBDTwZiogk:matrix.org/$3ZmHgbmQ6830krosfesFzpRw_nquUQ8XZte5jSilAes?via=libera.chat&via=matrix.org&via=ralith.com). See also previous discussion in https://github.com/rust-windowing/raw-window-handle/issues/59, and the related issue in https://github.com/rust-windowing/raw-window-handle/issues/85.

So I've opened this PR to see what it would take to make the handles `Send + Sync`, which would be somewhat possible on macOS, iOS and the web (provided that users uphold the "only use this on the main thread" requirement, which one could argue might be a safety footgun) - though I'm unsure if it's actually safe on the other platforms?

In any case we need to do _something_, because `winit::Window` is `Send + Sync`, and it unconditionally implements `HasWindowHandle`, so users could always just send the window to another thread and get a handle from that.

Fixes https://github.com/rust-windowing/raw-window-handle/issues/85 and fixes https://github.com/rust-windowing/raw-window-handle/issues/147.

TODO:
- [ ] Changelog entry
- [ ] Fix todos about safety comments

### DIFF
--- a/src/android.rs
+++ b/src/android.rs
@@ -29,6 +29,10 @@ pub struct AndroidNdkWindowHandle {
     pub a_native_window: NonNull<c_void>,
 }
 
+// SAFETY: `ndk::native_window::NativeWindow` is `Send + Sync`.
+unsafe impl Send for AndroidNdkWindowHandle {}
+unsafe impl Sync for AndroidNdkWindowHandle {}
+
 impl AndroidNdkWindowHandle {
     /// Create a new handle to an `ANativeWindow`.
     ///
--- a/src/appkit.rs
+++ b/src/appkit.rs
@@ -22,12 +22,26 @@ impl AppKitDisplayHandle {
 }
 
 /// Raw window handle for AppKit.
+///
+/// Note that while this is `Send + Sync`, it is only usable from the main
+/// thread. Any usage of `NSView` outside the main thread may be undefined
+/// behaviour, unless explicitly documented otherwise.
+///
+/// You must check whether the thread is the main thread before accessing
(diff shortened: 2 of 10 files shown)


kchibisov:
We had a lengthy discussion with @daxpedda on matrix, and I believe we end up with a solution different to what you propose.

The thing is that we must not mark it as a `Send + Sync`, since it's a footgun, and in reality it's not a `Send` + `Sync` on all the platforms. We can mark individual handlers as `Send + Sync`, but we can't guarantee that the handler will be `Send + Sync` all the time.

>In any case we need to do something, because winit::Window is Send + Sync, and it unconditionally implements HasWindowHandle, so users could always just send the window to another thread and get a handle from that.

The solution to that was to get rid of the handler traits or split getting a handlers out of some pseudo struct (has no use other than give you handle based on thread). In the end, what winit will have is a regular `raw_window_handle` getter, but which will fail if you call to it
on a thread other than event loop's thread, and an `unsafe` method, which will work, but at least there will be an attempt to deliver what you need to do and it'll that, while you have handler, normal reading/writing won't work like on the main thread, and you'll need extra sync.

Also, making something `Send` and `Sync` when in reality they are not doesn't make any sense, because users can clearly just *wrap* the type and send across the thread boundaries.

PR #154, opened 2023-10-13
Make more handles `!Send` and `!Sync`, and document main thread safety

madsmtm:
Follow-up to https://github.com/rust-windowing/raw-window-handle/pull/152 to consider documenting our decision on main thread safety, and making some handles `!Send` and `!Sync`.


madsmtm:
> might not want to do this

Yet, at least ;)

src/web.rs
```diff
@@ -33,6 +34,7 @@ pub struct WebWindowHandle {
     ///
     /// [data attributes]: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/data-*
     pub id: u32,
+    _marker: PhantomData<*const ()>,
```
daxpedda, line 37: I'm not sure what to think of this, the ID here is not a pointer, but it might be misleading for consumers who think that they are on the main thread if they got this variant.

Is there any previous discussion on this?

daxpedda: I was also wondering if there are any plans to remove this type completely, now that we have the `WebCanvasWindowHandle` and `WebOffscreenCanvasWindowHandle` types.

madsmtm: > plans to remove this type completely

Moved to https://github.com/rust-windowing/raw-window-handle/issues/157.

madsmtm: > I'm not sure what to think of this, the ID here is not a pointer, but it might be misleading for consumers who think that they are on the main thread if they got this variant.

Well, that's the same as what I've documented on e.g. `AppKitWindowHandle`, but I'm beginning to see that that may not be the best approach (hence why this PR isn't merged yet).

daxpedda: After some more thinking I'm now in favor of making this `!Send`.

PR #160, opened 2024-02-22
Improve documentation for UIKit and AppKit handles

madsmtm:
This takes part of the main thread only docs from https://github.com/rust-windowing/raw-window-handle/pull/154, but with the less strong guarantee of the handle only ever _existing_ on the main thread (as that's probably not yet true in the wild, so I'd rather have consumers to the extra check).

### DIFF
--- a/src/appkit.rs
+++ b/src/appkit.rs
@@ -48,6 +48,41 @@ impl DisplayHandle<'static> {
 /// Note that `NSView` can only be accessed from the main thread of the
 /// application. This struct is `!Send` and `!Sync` to help with ensuring
 /// that.
+///
+/// # Example
+///
+/// Getting the view from a [`WindowHandle`][crate::WindowHandle].
+///
+/// ```no_run
+/// # fn inner() {
+/// #![cfg(target_os = "macos")]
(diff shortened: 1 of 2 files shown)


Lokathor (approved):

Lokathor:
Other than the wasm CI failing, which I think is unrelated, it's always good to have usage example docs.

madsmtm:
> wasm CI failing

Yeah, Winit had the same problem, not sure how `cargo hack` works so don't know how to fix it, so I'm just gonna merge it regardless.

Lokathor:
Do we have a "wasm expert" to call for things like this? If 
not, should we move the wasm jobs to a separate CI path and just let them fail when they fail?

madsmtm:
That would be @daxpedda

daxpedda:
There is no good fix for this, this is a Cargo problem: https://github.com/rust-lang/rfcs/pull/3537.

The best we can do here, which we have done in Winit as well, is pin the `bumpalo` version in CI:
https://github.com/rust-windowing/winit/blob/a127bd6f6684909852000cdf32469bac3b611887/.github/workflows/ci.yml#L92

PR #164, opened 2024-04-12
Add OpenHarmony OS support

jschwe:
Add support for the [OpenHarmony] operating system, which is a [tier-2 target] in Rust starting with
Rust 1.78.
On OpenHarmony [ArkUI] is used to develop
applications. ArkUI provides the [XComponent]
type, which allows native code (e.g. Rust) to use
EGL / OpenGL ES to render directly to a native
window.

[tier-2 target]: https://doc.rust-lang.org/nightly/rustc/platform-support/openharmony.html
[OpenHarmony]: https://gitee.com/openharmony/docs/blob/master/en/OpenHarmony-Overview.md
[ArkUI]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/ui/arkui-overview.md
[XComponent]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/ui/arkts-common-components-xcomponent.md

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -135,7 +135,7 @@ pub enum RawWindowHandle {
     /// A raw window handle for the OpenHarmony OS NDK
     ///
     /// ## Availability Hints
-    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
+    /// This variant is used on OpenHarmony OS (`target_env = "ohos"`).
     OhosNdk(OhosNdkWindowHandle),
     /// A raw window handle for Xlib.
     ///
@@ -288,7 +288,7 @@ pub enum RawDisplayHandle {
     /// A raw display handle for OpenHarmony OS NDK
     ///
     /// ## Availability Hints
-    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
+    /// This variant is used on OpenHarmony OS (`target_env = "ohos"`).
     Ohos(OhosDisplayHandle),
     /// A raw display handle for Xlib.
     ///
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -37,6 +37,7 @@ mod android;
 mod appkit;
 mod borrowed;
 mod haiku;
+mod ohos;
 mod redox;
 mod uikit;
 mod unix;
@@ -47,6 +48,7 @@ pub use android::{AndroidDisplayHandle, AndroidNdkWindowHandle};
 pub use appkit::{AppKitDisplayHandle, AppKitWindowHandle};
 pub use borrowed::{DisplayHandle, HasDisplayHandle, HasWindowHandle, WindowHandle};
 pub use haiku::{HaikuDisplayHandle, HaikuWindowHandle};
+pub use ohos::{OhosDisplayHandle, OhosNdkWindowHandle};
 pub use redox::{OrbitalDisplayHandle, OrbitalWindowHandle};
 pub use uikit::{UiKitDisplayHandle, UiKitWindowHandle};
 pub use unix::{
@@ -130,6 +132,11 @@ pub enum RawWindowHandle {
     /// This variant is used by the Orbital Windowing System in the Redox
     /// operating system.
     Orbital(OrbitalWindowHandle),
+    /// A raw window handle for the OpenHarmony OS NDK
+    ///
+    /// ## Availability Hints
+    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
+    OhosNdk(OhosNdkWindowHandle),
     /// A raw window handle for Xlib.
     ///
     /// ## Availability Hints
@@ -278,6 +285,11 @@ pub enum RawDisplayHandle {
     /// This variant is used by the Orbital Windowing System in the Redox
     /// operating system.
     Orbital(OrbitalDisplayHandle),
+    /// A raw display handle for OpenHarmony OS NDK
+    ///
+    /// ## Availability Hints
+    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
+    Ohos(OhosDisplayHandle),
     /// A raw display handle for Xlib.
     ///
     /// ## Availability Hints
@@ -391,6 +403,7 @@ macro_rules! from_impl {
 from_impl!(RawDisplayHandle, UiKit, UiKitDisplayHandle);
 from_impl!(RawDisplayHandle, AppKit, AppKitDisplayHandle);
 from_impl!(RawDisplayHandle, Orbital, OrbitalDisplayHandle);
+from_impl!(RawDisplayHandle, Ohos, OhosDisplayHandle);
 from_impl!(RawDisplayHandle, Xlib, XlibDisplayHandle);
 from_impl!(RawDisplayHandle, Xcb, XcbDisplayHandle);
 from_impl!(RawDisplayHandle, Wayland, WaylandDisplayHandle);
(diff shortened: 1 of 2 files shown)


src/ohos.rs
```diff
@@ -0,0 +1,78 @@
...
+        // SAFETY: No data is borrowed.
+        unsafe { Self::borrow_raw(OhosDisplayHandle::new().into()) }
+    }
+}
+
+/// Raw window handle for Ohos NDK.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct OhosNdkWindowHandle {
+    pub xcomponent: NonNull<c_void>,
```
kchibisov, line 50: When you create `EGLDisplay` what you pass to it? I'd assume the `xcomponent`, but not a `native_window`? Or how does it work?

src/ohos.rs
```diff
@@ -0,0 +1,78 @@
...
+    pub native_window: NonNull<c_void>,
```
madsmtm, line 51: Maybe you could link to documentation about these?

And can you get one from the other? If so, then I'd strongly prefer that we only store one of these pointers.

notgull: Yes, please link documentation. This would be an entirely new target.

jschwe:
Sorry for the slow reply.

I believe I was mistaken with my previous statement that the `xcomponent` pointer is also required - at least for my usecase of getting [servo](https://github.com/servo/servo) running on  OpenHarmony, the APIs provided by [OHNativeWindow] seem to be sufficient.
I adjusted my PR and removed the `XComponent`, but since I'm not familiar with UIs and what exactly `RawWindowHandle`s can be used for on other systems, let me give a short introduction and link the relevant documentation.

The latest version of the XComponent Guidelines available in English, seems to be [Native XComponent Guidelines 4.0 (English)].
On the master branch there currently is only the [Chinese version][Native XComponent Guidelines master (Chinese)], but version 4.0 is quite recent too.

To summarize the workflow (with the C++ examples in the documentation):

1. The UI is defined in TypeScript code (ArkUI / ArkTS)
2. The [XComponent] with the surface type can be used to get a native EGL surface
3. Native code can receive a callback `OnSurfaceCreatedCB(OH_NativeXComponent *component, void *window)` when the XComponent is created.
4. The documentation suggests determining eglDisplay via `eglGetDisplay(EGL_DEFAULT_DISPLAY)`. In servo I've been using `egl.GetCurrentDisplay();` (same way as the native android servo app)
5. The `void* window` ([OHNativeWindow]  / `EGLNativeWindowType`) argument to OnSurfaceCreatedCB, is the eglWindow needed  to create an eglSurface with `eglCreateWindowSurface(eglDisplay_, eglConfig_, eglWindow_, NULL);

The [API provided by OHNativeWindow][OHNativeWindow] is quite low level, and missing the following APIs compared 
to the [XComponent] it belongs to:

- Get the offset of the XComponent relative to the upper left of the screen.
- Touch / Mouse / Key events
- Registering new Callbacks for the XComponent
- Set expected frame rate

I'm currently leaning towards it being fine if the `OhosNdkWindowHandle` would not allow usage of such APIs, but
I'm also not very familiar how `raw-window-handle` is used in practice (besides servo), so I'm very open to feedback.

[XComponent]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/apis-arkui/_o_h___native_x_component.md
[Native XComponent Guidelines master (Chinese)]: https://gitee.com/openharmony/docs/blob/master/zh-cn/application-dev/ui/napi-xcomponent-guidelines.md
[Native XComponent Guidelines 4.0 (English)]: https://gitee.com/openharmony/docs/blob/OpenHarmony-4.0-Release/en/application-dev/napi/xcomponent-guidelines.md
[Native Window Guidelines]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/graphics/native-window-guidelines.md
[OHNativeWindow]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/reference/apis-arkgraphics2d/_native_window.md

src/ohos.rs
jschwe: I added some documentation  to the module,  including links to the relevant OpenHarmony documentation pages.

> And can you get one from the other? If so, then I'd strongly prefer that we only store one of these pointers.

Not as far as I know.  The native window and the xcomponent belong together, and are received via a callback when the XComponent is created.

madsmtm:
I spent a few minutes trying to read the docs, and I still have no idea what an "XComponent" actually is? As in, what does it correspond to visually? Is it the top-level window itself? Or can components be nested like subsurfaces/subviews? If I tried to do some GUI development in OpenHarmony, and I wanted to place a button on the screen somewhere, would that button be a "subclass" (or similar) of XComponent, of OHNativeWindow, or of something entirely different?

jschwe:
On OpenHarmony [ArkTS] is the usual language apps including their UI are written in. ArkTS is a flavor of typescript, which adds some restrictions, like forcing initialization of variables and requiring that all types are known at compile time (i.e. `any` is forbidden). Additionally, ArkTS adds builtins for UI components ([example](https://gitee.com/openharmony/docs/blob/master/en/application-dev/quick-start/arkts-basic-syntax-overview.md)).

The [ArkTS XComponent](en/application-dev/ui/arkts-common-components-xcomponent.md) can be used in the ArkTS UI code to obtain a surface, that the native C++ (or in our case Rust) code can render to using OpenGL. The size and position of the XComponent is determined by the ArkTS UI layout.

[In this example](https://github.com/jschwe/ServoDemo/blob/001f3e868cf949a9f9904e859af44d892b56fe7c/entry/src/main/ets/pages/Index.ets#L25) the ArkTS UI definition is simply one Textbox (which is used as a URL bar) and beneath that an `XComponent`, which [servo](https://github.com/servo/servo) renders to using OpenGL.

```
Column() {
  TextInput(My_url_bar)
  XComponent()
}
```

See also [Introduction to ArkUI](https://gitee.com/openharmony/docs/blob/master/en/application-dev/ui/Readme-EN.md)

> If I tried to do some GUI development in OpenHarmony, and I wanted to place a button on the screen somewhere, would that button be a "subclass" (or similar) of XComponent, of OHNativeWindow, or of something entirely different?

You would use the [ArkTS Button builtin](https://gitee.com/openharmony/docs/blob/master/en/application-dev/ui/arkts-common-components-button.md). The XComponent (surface type) is used when rendering directly from native code via OpenGL.

[ArkTS]: https://gitee.com/openharmony/docs/blob/master/en/application-dev/quick-start/arkts-get-started.md

madsmtm:
Can you query the size of the surface from just `OHNativeWindow`?

src/lib.rs
```diff
@@ -130,6 +132,11 @@ pub enum RawWindowHandle {
     /// This variant is used by the Orbital Windowing System in the Redox
     /// operating system.
     Orbital(OrbitalWindowHandle),
+    /// A raw window handle for the OpenHarmony OS NDK
+    ///
+    /// ## Availability Hints
+    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
```
madsmtm, line 138: Nit:

```suggestion
    /// This variant is used on OpenHarmony OS (`target_env = "ohos"`).
```

src/lib.rs
```diff
@@ -278,6 +285,11 @@ pub enum RawDisplayHandle {
     /// This variant is used by the Orbital Windowing System in the Redox
     /// operating system.
     Orbital(OrbitalDisplayHandle),
+    /// A raw display handle for OpenHarmony OS NDK
+    ///
+    /// ## Availability Hints
+    /// This variant is used on OpenHarmony OS ( `target_env = "ohos"`).
```
madsmtm, line 291:
```suggestion
    /// This variant is used on OpenHarmony OS (`target_env = "ohos"`).
```

madsmtm (approved):
Cool, then I think this makes sense as-is, and if we later find that the XComponent is actually needed, then we can always add an optional field.

jschwe:
Is there anything I still need to do before this can be merged? Should I squash?

Lokathor:
I can squash-merge

PR #168, opened 2024-05-17
Update Changelog and bump version number

jschwe:
It would be nice (for me) if there could be a new Release containing #164.
I just went ahead and put todays date in the Changelog, but I can also change it to `## unreleased` or something else.


Lokathor:
0.6.2 should be available now

PR #174, opened 2025-01-26
Improve documentation on AppKit and UIKit variants 

madsmtm:
- Update availability hints to include visionOS.
- Use objc2 v0.6 in examples.
- Add more detailed examples of how to create UIKit and AppKit handles.


notgull (approved):
Thanks!

PR #175, opened 2025-01-26
Deprecate `UiKitWindowHandle::ui_view_controller`

madsmtm:
This can be retrieved from the `UIView` itself, and is hence unnecessary information for us to carry around.

Besides, since it's optional, users that want to get the view controller has to do the searching logic anyways.

Once we remove this in the next breaking version, we should move the example code I've provided on `ui_view_controller` here to an example instead.


notgull (approved):
Thanks!

PR #176, opened 2025-10-05
Add UEFI support

RossComputerGuy:


src/lib.rs
```diff
@@ -207,6 +215,12 @@ pub enum RawWindowHandle {
     /// ## Availability Hints
     /// This variant is used on HaikuOS.
     Haiku(HaikuWindowHandle),
+    /// A window handle for UEFI.
+    ///
+    /// ## Availability Hints
+    /// This variant is used on UEFI.
+    #[cfg(target_os = "uefi")]
```
notgull, line 222:
```suggestion
```
We avoid target conditionals here.

src/lib.rs
```diff
@@ -414,6 +434,9 @@ from_impl!(RawDisplayHandle, Web, WebDisplayHandle);
 from_impl!(RawDisplayHandle, Android, AndroidDisplayHandle);
 from_impl!(RawDisplayHandle, Haiku, HaikuDisplayHandle);
 
+#[cfg(target_os = "uefi")]
```
notgull, line 437:
```suggestion
```

Ditto

src/uefi.rs
```diff
@@ -0,0 +1,59 @@
...
+    /// let handle = DisplayHandle::uefi();
+    /// do_something(handle);
+    /// ```
+    pub fn uefi() -> Result<Self, HandleError> {
+        let proto = boot::get_handle_for_protocol::<GraphicsOutput>()
+            .or_else(|_| Err(HandleError::Unavailable))?;
+        // SAFETY: No data is borrowed.
+        Ok(unsafe { Self::borrow_raw(UefiDisplayHandle::new(proto).into()) })
+    }
+}
```
notgull, line 48:
```suggestion
```

This level of functionality is too high for this crate.

RossComputerGuy:
> * From a quick glance at the docs, it seems that `GraphicsOutput` has a framebuffer of its own. So I think it would be better if `GraphicsOutput` was a property of the `WindowHandle` rather than the `DisplayHandle`.

The framebuffer is only available when the Graphics Output Protocol implements a framebuffer mode. Otherwise, you have to use blit mode.

notgull:
Still, you draw to the `GraphicsOutput`, which in line with our earlier implementation for DRM/KMS. I think if you draw to it, it should be classified as a window.

RossComputerGuy:
I'm not sure what you mean. When drawing in UEFI, you use a temporary buffer which you render to. Then whenever you present, you copy from that buffer to the memory mapped one if it is available. Otherwise, you use the blit function in the protocol. If the memory mapped buffer is not available, that means the GPU just doesn't give the ability for direct screen access.

I was wanting to use this with something like iced and have it like a compositor. Have multiple windows which get drawn and be able to manipulate them. At least in this crate and softbuffer, it shouldn't be anything special to make that possible.

src/lib.rs
RossComputerGuy: Done

src/uefi.rs
RossComputerGuy: Done

src/uefi.rs
```diff
@@ -0,0 +1,40 @@
+use core::ffi::c_void;
+use core::ptr::NonNull;
+
+/// Raw display handle for UEFI.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct UefiDisplayHandle {
+    /// The UEFI Graphics Output Protocol handle.
+    pub handle: NonNull<c_void>,
```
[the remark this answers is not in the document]
RossComputerGuy: Done, I clarified that it comes from [spec v2.0](https://uefi.org/sites/default/files/resources/UEFI_Specification_2_and_Errata_Sept16_08.pdf) or newer under `EFI_GRAPHICS_OUTPUT_PROTOCOL`.

src/uefi.rs
```diff
@@ -0,0 +1,43 @@
...
+use core::ptr::NonNull;
+
+/// Raw display handle for UEFI.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+pub struct UefiDisplayHandle {
+    /// The UEFI Graphics Output Protocol handle.
+    ///
+    /// From the UEFI spec on <https://uefi.org>, can be found in versions since 2.0 as
+    /// `EFI_GRAPHICS_OUTPUT_PROTOCOL`.
```
madsmtm:
```suggestion
    /// [`EFI_GRAPHICS_OUTPUT_PROTOCOL`].
    ///
    /// [`EFI_GRAPHICS_OUTPUT_PROTOCOL`]: https://uefi.org/specs/UEFI/2.10/12_Protocols_Console_Support.html#efi-graphics-output-protocol
```

Right?

madsmtm:
Re-reading the docs for `RawDisplayHandle`, it says:
> The display usually represents a connection to some display server, but it is not necessarily tied to a particular window. Some APIs can use the display handle without ever creating a window handle (e.g. offscreen rendering, headless event handling).

Which seems to indicate that perhaps display handles are better thought of as "connection handle" (they don't actually correspond to a physical "display"/monitor).

Idk., the analogies break down somewhat at this point. Maybe the best approach is to consider how things will be used? Maybe you could expand on how you produce this handle for `softbuffer` to consume it?

madsmtm:
This is blocked on https://github.com/rust-windowing/softbuffer/pull/282#pullrequestreview-3718339199 (I'm still not convinced that this is better represented as a window/"surface" handles).

Issue #171, opened 2024-07-14
Change Windows `HANDLE` types back to `*mut c_void` **again** :weary:?


MarijnS95:
> FYI, `win32metadata` is changing these back to void pointers it seems:
>
> https://github.com/microsoft/win32metadata/issues/1924
> https://github.com/microsoft/win32metadata/commit/b4dfd2f25c0114da8b27d0d390bca5c1555fc55c
>
> See also my local regeneration: https://github.com/microsoft/windows-rs/commit/96b9a27a6293fe08e4a96b64400239594941ff20

_Originally posted by @MarijnS95 in https://github.com/rust-windowing/raw-window-handle/pull/136#discussion_r1640368547_

---

This is out now in the `windows 0.58` release and somewhat annoying when using `raw-window-handle` together with it.

Lokathor:
*deep and heavy sigh*

madsmtm:
This should be combined with https://github.com/rust-windowing/raw-window-handle/pull/154

notgull:
We can change them back to pointers. If they decide to go back to `usize` just keep them as pointers.

notgull:
Actually wait, does this impact us? From the issue linked above:

> Rust currently emits isize/usize integers as a result of decomposing our synthetic handle types and encountering IntPtr/UIntPtr. Rust would now like to emit pointers to better align with its core handle type, but it cannot differentiate between legitimate pointer-sized integers (e.g. LRESULT) and handles (e.g. HMODULE).

Since window IDs in Windows are just indexes into a thread-local table, they're still semantically `usize`/`isize`. So it's still correct (by Rust's standards) to not change them here, even if `win32metadata` doesn't realize it.

PR #183, opened 2026-02-07
breaking: Change Win32 to use pointers instead of ints

notgull:
Needed as per new Win32 metadata updates

Fixes: #171
Supersedes: #154

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -479,7 +479,7 @@ mod tests {
         assert_not_impl_any!(WaylandWindowHandle: Send, Sync);
         assert_impl_all!(DrmWindowHandle: Send, Sync);
         assert_not_impl_any!(GbmWindowHandle: Send, Sync);
-        assert_impl_all!(Win32WindowHandle: Send, Sync);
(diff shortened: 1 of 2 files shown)


madsmtm:
> Or maybe provide method to get a `ptr` or maybe store as ptr and get a method to get as usize. We were storing them as ptr before, then they were made into integer per migration to windows-rs stuff, and now we move them back to ptr, so at this point maybe just having both should be fine?

I'm against _storing_ both, but we could have a method to cast these.

Then again, users on all other platforms are gonna need a cast from `c_void` to whatever , it's not really a big deal for a `windows-sys v0.52.0` user to do that cast themselves.

notgull:
> Or maybe provide method to get a `ptr` or maybe store as ptr and get a method to get as usize. We were storing them as ptr before, then they were made into integer per migration to windows-rs stuff, and now we move them back to ptr, so at this point maybe just having both should be fine?

Not sure if this is useful. `windows-sys` currently represents the handle as a pointer. If it goes back to being an integer for some reason, we can add helper methods for converting to and from integer.

Issue #177, opened 2026-02-06
Stability


notgull:
[SemVer](https://semver.org/) says "If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backward compatibility, you should probably already be 1.0.0." Given that `raw-window-handle` is:

- Used in production as a fundamental crate for many GUI applications.
- Has a stable API that dozens of GUI crates expose publicly.
- Has frequent panics about backwards compatibility.

...then we probably should release 1.0.

Given that the current API allows for 100% safety and has been in use for 2.5 years with few complaints (documented [here](https://github.com/rust-windowing/raw-window-handle/milestone/1)), I feel as though we're in a good position to release `raw-window-handle` v1.0.

Small checklist of items:

- [x] Change window types to use pointers again. (#171)
- [x] ~~Make all handles `!Send` and `!Sync`. (#154)~~ Not planned.
- [x] ~~Determine best course of action for web handles. (#185)~~ Current approach works for now.
- [ ] Determine if it's better to have a handle for a "view" as opposed to a "window". (#123)
- [x] Should we merge Display handles on platforms where they're always available? (#180)
- [ ] Should we get rid of `RawWindowHandle`, and merge it all into `WindowHandle<'_>`? (#178)
- [x] Variance of window handles. (#179)
- [ ] Implement owned window handles. (#188)
- [ ] Implement a C API. (#189)

notgull:
cc @rust-windowing/winit

Lokathor:
I agree we should move to a 1.0

madsmtm:
I agree that we should move towards this.

---

I'm still against this library having the `wasm-bindgen-0-2` feature flag introduced in https://github.com/rust-windowing/raw-window-handle/pull/134 - I feel like the helpers introduced there, while it does allow the user to avoid `unsafe`, it doesn't significantly help IMO unless we do the same thing on every backend, and recommend that as the approach. I.e. have a `objc2-0-6` feature flag, and a `windows-0-61` flag.

As a data point, I tried looking at the reverse dependencies on crates.io, and found that **not a single dependency uses the `wasm-bindgen-0-2` feature flag**.

I feel that the approach I've gone with for the AppKit and UIKit handles, i.e. [documenting the recommended way to construct and use them](https://docs.rs/raw-window-handle/0.6.2/raw_window_handle/struct.AppKitWindowHandle.html), is the better way to go.

[...]

PR #184, opened 2026-02-07
breaking: Remove wasm-bindgen from public API

notgull:
This commit replaces the `JsValue` in the public API
with the underlying index, as per earlier discussion.

Ref: #177

### DIFF
--- a/src/web.rs
+++ b/src/web.rs
@@ -96,11 +96,24 @@ impl WebCanvasWindowHandle {
     ///
     /// # Example
     ///
-    /// ```
+    #[cfg_attr(target_family = "wasm", doc = "```no_run")]
+    #[cfg_attr(not(target_family = "wasm"), doc = "```no_compile")]
     /// # use raw_window_handle::WebCanvasWindowHandle;
-    /// let value: usize;
-    /// # value = 0;
-    /// let mut handle = WebCanvasWindowHandle::new(value);
+    /// use core::mem::ManuallyDrop;
+    /// use wasm_bindgen::convert::{IntoWasmAbi, RefFromWasmAbi};
+    /// use web_sys::HtmlCanvasElement;
+    ///
+    /// let value: HtmlCanvasElement;
+    /// # value = todo!();
+    ///
+    /// // Convert to the raw index and convert to the handle.
+    /// let index = (&value).into_abi();
+    /// let mut handle = WebCanvasWindowHandle::new(index as usize);
+    ///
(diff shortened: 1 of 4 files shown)


src/web.rs
```diff
@@ -106,45 +97,16 @@ impl WebCanvasWindowHandle {
...
-
-    /// Convert to the underlying [`wasm_bindgen::JsValue`].
-    ///
-    /// # Safety
-    ///
-    /// The inner pointer must be valid. This is ensured if this handle was
-    /// borrowed from [`WindowHandle`][crate::WindowHandle].
-    pub unsafe fn as_wasm_bindgen_0_2(&self) -> &wasm_bindgen::JsValue {
-        unsafe { self.obj.cast().as_ref() }
+    pub fn new(obj: usize) -> Self {
```
madsmtm: Is this safe? Could we get a comment explaining why?

daxpedda (changes requested):
I actually don't know where it was discussed to switch to indices from pointers. Its a very interesting path forward though that I approve of.

I spent some time right now trying to figure out what would exactly be necessary to transfer across toolchains, but I will lay it out in more detail in #185.

Until we get there however, the safety of this API doesn't change. We still have to expect that on both ends we are talking about a `wasm_bindgen::JsValue`. So for the input side we have to guarantee that it stay alive as long as the handle is valid. On the output side we have to guarantee to not `Drop` it. This is very important otherwise the actual `externref` would be cleaned up.

The old API expected a `*const JsValue` input and a `&JsValue` output. Instead we now have to use [`IntoWasmAbi`](https://docs.rs/wasm-bindgen/0.2.108/wasm_bindgen/convert/trait.IntoWasmAbi.html) for the input and [`RefFromWasmAbi`](https://docs.rs/wasm-bindgen/0.2.108/wasm_bindgen/convert/trait.RefFromWasmAbi.html) for the output. We should most definitely provide an example for this, otherwise its quite hard to get it right.

notgull:
> The old API expected a `*const JsValue` input and a `&JsValue` output. Instead we now have to use [`IntoWasmAbi`](https://docs.rs/wasm-bindgen/0.2.108/wasm_bindgen/convert/trait.IntoWasmAbi.html) for the input and [`RefFromWasmAbi`](https://docs.rs/wasm-bindgen/0.2.108/wasm_bindgen/convert/trait.RefFromWasmAbi.html) for the output. We should most definitely provide an example for this, otherwise its quite hard to get it right.

Added this

Issue #157, opened 2023-12-22
Remove `WebWindowHandle`?


madsmtm:
We now have `WebCanvasWindowHandle` and `WebOffscreenCanvasWindowHandle`, which are much safer, so we should probably work on removing `WebWindowHandle`.

Probably with a deprecation period first.

daxpedda:
I'm in favor of this as well.

I've implemented support for the other handles in some known crates with success so far:
- https://github.com/rust-windowing/winit/pull/3270
- https://github.com/gfx-rs/wgpu/pull/4888
- https://github.com/rust-windowing/softbuffer/pull/188

notgull:
I think we should wait to remove these handles until we have a stable (post-1.0) way to refer to web handles, whether it's `wasm-bindgen` or the new `wit-bindgen` thing that's coming out.

PR #186, opened 2026-02-11
breaking: Remove web window handle

notgull:
This handle type is no longer used by any windowing systems
and only exists as a burden for graphics implementors.

Fixes: #157


madsmtm (approved):
Maybe add a changelog entry?

PR #187, opened 2026-02-14
Add a handle for views

notgull:
This MR adds a handle type specifically for views. "Views" are defined
as sub-areas of windows that graphics APIs can render into. This is
separated from Windows, which may be a key difference for some APIs.

So far this MR only implements views that wrap around windows, and views
that wrap around NSView from AppKit. Discussion must be done to
determine if this is the best way to go about this, as well as if we
should add other views.

Ref: #123

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,10 @@
 ## Unreleased
 
 * **Breaking:** Rework web handles to remove `wasm-bindgen` from public API. (#184)
-- **Breaking:** Remove `WebWindowHandle` as it is no longer used. (#186)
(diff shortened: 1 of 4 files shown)


notgull:
AI: Change name to `SurfaceHandle`.

notgull:
I'm actually wondering if this isn't the right play here. Maybe just having a macOS window handle with a back-pointer to the `NSWindow` would work, just like how we are currently planning to do Wayland windows.

```rust
struct AppkitWindowHandle {
    ns_view: NonNull<c_void>,
    ns_window: Option<NonNull<c_void>>
}
```

PR #192, opened 2026-02-22
breaking: Change thread safety guarantees for window types

notgull:
In this commit, thread safety documentation is added for all window
handle types. In addition, following some additional research, some
window types have their thread safety changed. Most notably:

- OHOS windows are definitely thread-unsafe.
- Redox and Haiku windows are actually thread safe.
- GBM types are Send but not Sync.
- Win32 windows are Sync but not Send.
- Display handles are changed to match their windows.

### DIFF
--- a/src/appkit.rs
+++ b/src/appkit.rs
@@ -1,12 +1,23 @@
 use core::ffi::c_void;
+use core::marker::PhantomData;
 use core::ptr::NonNull;
 
 use super::DisplayHandle;
 
 /// Raw display handle for AppKit.
+///
+/// ## Thread Safety
+///
+/// This type has the safe thread safety guarantees as [`AppKitWindowHandle`].
+///
+/// Note that this type does not contain any Appkit objects. However,
+/// it is kept `!Send` and `!Sync` for the event that Appkit objects are
+/// added to this type.
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-pub struct AppKitDisplayHandle {}
+pub struct AppKitDisplayHandle {
+    _thread_unsafe: PhantomData<*mut ()>,
+}
 
 impl AppKitDisplayHandle {
     /// Create a new empty display handle.
@@ -19,7 +30,9 @@ impl AppKitDisplayHandle {
     /// let handle = AppKitDisplayHandle::new();
     /// ```
     pub fn new() -> Self {
-        Self {}
+        Self {
+            _thread_unsafe: PhantomData,
+        }
     }
 }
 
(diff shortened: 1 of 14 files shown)


src/unix.rs
```diff
@@ -3,6 +3,14 @@ use core::num::NonZeroU32;
...
 
 /// Raw display handle for Xlib.
+///
+/// ## Thread Safety
+///
+/// Reads and writes to and from the X server are internally secure by a [mutex].
+/// Therefore this type is `Send` and `Sync. This means it can be sent to or
+/// used from other threads.
+///
+/// [mutex]: https://gitlab.freedesktop.org/xorg/lib/libxcb/-/blob/master/src/xcb_conn.c?ref_type=heads#L165
```
kchibisov: Given that we use `xlib` and not xcb for this handle it should point to xlib docs, which is also safe.

kchibisov (changes requested):

src/appkit.rs
```diff
@@ -83,6 +105,19 @@ impl DisplayHandle<'static> {
 /// }
 /// # }
 /// ```
+///
+/// ## Thread Safety
```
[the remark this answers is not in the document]
notgull: Where would these texts be merged?

src/web.rs
```diff
@@ -45,6 +63,12 @@ impl DisplayHandle<'static> {
 /// Raw window handle for a Web canvas registered via [`wasm-bindgen`].
 ///
 /// [`wasm-bindgen`]: https://crates.io/crates/wasm-bindgen
+///
+/// ## Thread-Safety
+///
+/// WASM objects are usually bound to the main UI "thread" belonging to the
+/// top-level webpage. Therefore this type is `!Send` and `!Sync`. It cannot be
+/// sent to or used from other threads.
```
madsmtm: CC @daxpedda on this.

src/windows.rs
```diff
@@ -89,13 +112,28 @@ impl Win32WindowHandle {
...
+/// functions especially), so care should be taken to not call these functions
+/// from other threads. When in doubt, only run the function on the main thread.
+///
+/// [thread affinity]: https://devblogs.microsoft.com/oldnewthing/20051010-09/?p=33843
 #[non_exhaustive]
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
 pub struct WinRtWindowHandle {
     /// A WinRT `CoreWindow` handle.
     pub core_window: NonNull<c_void>,
 }
```
madsmtm: Do you have docs that this really applies to WinRT too? I'd be worried that it might be more restrictive than Win32.

madsmtm (commented):
I've looked at the backends I could, generally think this PR is a good change. We might want to refine the Windows stuff, but let's discuss that in https://github.com/rust-windowing/raw-window-handle/pull/209.

src/web.rs
daxpedda:
```suggestion
/// JS objects can't be implicitly transferred across threads.
/// Therefore this type is `!Send` and `!Sync`. It cannot be
/// sent to or used from other threads.
```

Issue #193, opened 2026-02-28
Improve documentation on how producer and consumer crates are supposed to act


madsmtm:
We talk a lot about producer crates like `winit`/`sdl`/`wry`/... and consumer crates like `wgpu`/`softbuffer`/`muda`/... etc., but we don't actually mention this distinction in our documentation.

We should document this better, definitely in `lib.rs`, perhaps also with a bigger example.

It might also make sense to maintain a list of producers and consumers in our docs?

PR #202, opened 2026-02-28
Document producer/consumer relationship

madsmtm:
Part of https://github.com/rust-windowing/raw-window-handle/issues/193.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,7 +7,8 @@
 //!
 //! This crate is an interoperability interface that allows "producer" crates to create a handle to
 //! a window, and for this handle to be passed onwards to "consumer" crates for rendering to the
-//! window's surface.
+//! window's surface. This solves the MxN integration problem: consumer and producer crates don't
+//! have to know about / depend on each other.
 //!
 //! This crate does not by itself provide any utilities for creating and managing windows, nor for
 //! rendering into them.


src/lib.rs
```diff
@@ -3,12 +3,55 @@
...
+//! "reference implementations" of consumer crates. Other examples include
+//! [`ash-window`](https://docs.rs/ash-window/) and [`pixels`](https://docs.rs/pixels/).
+//!
+//! Another type of producer is those that use the window handle to access either the platform
+//! widget/view or the actual window, and performs some operation that requires that. Examples of
+//! this include [`rfd`](https://docs.rs/rfd/), [`wry`](https://docs.rs/wry/),
+//! [`muda`](https://docs.rs/muda/), [`window_clipboard`](https://docs.rs/window_clipboard/),
+//! [`ashpd`](https://docs.rs/ashpd/), [`window-vibrancy`](https://docs.rs/window-vibrancy/),
+//! [`window-shadows`](https://docs.rs/window-shadows/) and
+//! [`native-dialog`](https://docs.rs/native-dialog/).
```
MarijnS95: These are **consumers**, right? They _take_ a RWH to interface with the OS for non-graphics like clipboard access, popup dialogs and whatnot (perhaps mention what these crates do in brief, as it's not always clear from the crate name).

```suggestion
//! Another type of consumer is those that use the window handle to access either the platform
//! widget/view or the actual window, and performs some operation that requires that. Examples of
//! this include [`rfd`](https://docs.rs/rfd/), [`wry`](https://docs.rs/wry/),
//! [`muda`](https://docs.rs/muda/), [`window_clipboard`](https://docs.rs/window_clipboard/),
//! [`ashpd`](https://docs.rs/ashpd/), [`window-vibrancy`](https://docs.rs/window-vibrancy/),
//! [`window-shadows`](https://docs.rs/window-shadows/) and
//! [`native-dialog`](https://docs.rs/native-dialog/).
```

src/lib.rs
```diff
@@ -3,12 +3,55 @@
 #![allow(clippy::new_without_default)]
 #![deny(unsafe_op_in_unsafe_fn)]
 
-//! Interoperability library for Rust Windowing applications.
+//! # Interoperability helper between graphics crates
```
daxpedda: I believe the new title is actually less accurate than the previous one.
What we offer here is interoperability between windowing and graphic libraries.

Potentially we also help multiple graphics libraries communicate a window between each other, but that's not really the main use case.

```suggestion
//! # Interoperability Types For Window Handles
```

src/lib.rs
```diff
@@ -3,12 +3,55 @@
...
+//! primary example of this is [`winit`](https://docs.rs/winit/), consider that the "reference
+//! implementation" of a producer crate.
+//!
+//! Other cross-platform producer crates include [`sdl3`](https://docs.rs/sdl3/),
+//! [`sdl2`](https://docs.rs/sdl2/), [`minifb`](https://docs.rs/minifb/),
+//! [`glfw`](https://docs.rs/glfw/) and [`fltk`](https://docs.rs/fltk/).
+//!
+//! Some platform-specific toolkits might also decide to implement `HasWindowHandle` for their
+//! types to allow more easily using them with consumer crates. Examples of this include `x11rb`,
+//! `ndk`, `orbclient` and `wayland-backend`.
```
daxpedda:
```suggestion
//! types to allow more easily using them with consumer crates. Examples of this include [`x11rb`](https://docs.rs/x11rb),
//! [`ndk`](https://docs.rs/ndk), [`orbclient`](https://docs.rs/orbclient) and [`wayland-backend`](https://docs.rs/wayland-backend).
```

src/lib.rs
```diff
@@ -3,12 +3,55 @@
...
+//!
+//! ## Consumers
+//!
+//! Consumer crates expose some sort of "surface" type that renders into a provided window handle.
```
daxpedda: I believe that in general we should reframe this as just an example of what a consumer crate might do. You already point out further down that there are consumers who just work with the window directly.

```suggestion
//! Consumer crates typically expose some sort of "surface" type that renders into a provided window handle.
```

src/lib.rs
madsmtm: Ofc., I wrote wrong

src/lib.rs
```diff
@@ -3,12 +3,55 @@
...
+//! ```
+//! struct Surface<W: HasWindowHandle> {
+//!     handle: W,
+//! }
+//! ```
+//!
+//! Examples of consumer crates include [`softbuffer`](https://docs.rs/softbuffer/),
+//! [`glutin`](https://docs.rs/glutin/) and [`wgpu`](https://docs.rs/wgpu/), consider these
+//! "reference implementations" of consumer crates. Other examples include
+//! [`ash-window`](https://docs.rs/ash-window/) and [`pixels`](https://docs.rs/pixels/).
```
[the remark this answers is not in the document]
MarijnS95: I do PR to `wgpu` from time to time (specifically WRT windowing / surface handles), but it's so humongous (and has an insane amount of insultingly low-quality PR contributions, I feel sorry for the maintainers) that it's almost impossible to get back into it every time, for example once one of the maintainers finally comes around to re-review my PR (because of otherwise having to deal with noise...).

---

`ash-window` is my area fwiw, but it can't do much if anything for lifetimes after handing out a cheaply-copyable `VkSurfaceKHR` handle; unless we re-wrap it in a smarter type that sparingly hands it out.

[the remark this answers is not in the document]
madsmtm: I've changed it to be a simple list instead.

src/lib.rs
```diff
@@ -3,12 +3,67 @@
...
+//! to the window's surface.
+//!
+//! This crate does not by itself provide any utilities for creating and managing windows, nor for
+//! rendering into them.
+//!
+//! ## Producers
+//!
+//! Producer crates provide some sort of "window" type that implements [`HasWindowHandle`], and
+//! ensures that the [`WindowHandle`] returned from that is alive for as long as the window is.
+//! Examples of cross-platform producer crates include:
```
notgull:
```suggestion
//! Examples of cross-platform producer crates include:
//! 
```

src/lib.rs
```diff
@@ -3,12 +3,67 @@
...
+//! Some platform-specific toolkits might also decide to implement `HasWindowHandle` for their
+//! types to allow more easily using them with consumer crates. Examples of this include:
+//! - [`x11rb`](https://docs.rs/x11rb).
+//! - [`wayland-backend`](https://docs.rs/wayland-backend).
+//! - [`ndk`](https://docs.rs/ndk).
+//! - [`orbclient`](https://docs.rs/orbclient).
+//!
+//! ## Consumers
+//!
+//! Consumer crates typically expose some sort of "surface" type that renders into a provided window
```
notgull:
```suggestion
//! Consumer crates expose a "surface" type that renders into a provided window
```

Strunk & White, Rule 17

src/lib.rs
```diff
@@ -3,12 +3,67 @@
...
+//! Examples of consumer crates include:
+//! - [`softbuffer`](https://docs.rs/softbuffer/) (surface for software rendering).
+//! - [`glutin`](https://docs.rs/glutin/) (OpenGL surface and context setup).
+//! - [`wgpu`](https://docs.rs/wgpu/) (general GPU interface).
+//! - [`ash-window`](https://docs.rs/ash-window/) (Vulkan API).
+//! - [`pixels`](https://docs.rs/pixels/) (software rendering built upon `wgpu`).
+//!
+//! Another type of consumer is those that use the window handle to access either the platform
+//! widget/view or the actual window, and performs some operation that requires that. Examples of
+//! this include:
```
notgull:
```suggestion
//! this pattern include:
//! 
```

src/lib.rs
madsmtm: See https://github.com/rust-windowing/raw-window-handle/pull/202#discussion_r2868874497, the "typically" is somewhat load-bearing.

src/lib.rs
```diff
@@ -3,12 +3,67 @@
...
-//! This library provides standard types for accessing a window's platform-specific raw window
-//! handle and platforms display handle. This does not provide any utilities for creating and
-//! managing windows; instead, it provides a common interface that window creation libraries (e.g.
-//! Winit, SDL) can use to easily talk with graphics libraries (e.g. gfx-hal).
+//! This crate is intended as an interoperability crate, that allows "producer" crates to create a
+//! handle to a window, and for this handle to be passed onwards to "consumer" crates for rendering
+//! to the window's surface.
+//!
+//! This crate does not by itself provide any utilities for creating and managing windows, nor for
+//! rendering into them.
```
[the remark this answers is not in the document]
notgull: ACK

Issue #185, opened 2026-02-07
Improve Toolchain Neutrality for Wasm


daxpedda:
Currently, we are already quite toolchain neutral by expecting an index to the `externref` table that holds the `HTMLCanvasElement` or `OffscreenCanvas`. This is going to cover every potential toolchain unless Wasm `externref` changes radically somehow.

Even if Rust introduces [first-class support for `externref`](https://github.com/rust-lang/rfcs/pull/3729), we still want an index to the table instead of the `externref` directly. This is because `externref` itself would be `!Sized` and therefor not fit into `RawWindowHandle`.

The Wasm [Shared Everything Threads Proposal](https://github.com/WebAssembly/shared-everything-threads) introduces shared tables, implying `Send + Sync`. However, this would not affect us unless `OffscreenCanvas` would also become a shared JS struct, which is unlikely nor is there any proposal of that sort. (`HTMLCanvasElement` will never be a shared JS struct because its DOM)

I believe that we should also add another index for the `externref` table. This would be relevant for toolchains with multiple `externref` tables. While I'm not aware of one, this would make sure we really stay as toolchain neutral as possible. I was thinking of this because [`js-bindgen`](https://github.com/wasm-bindgen/js-bindgen) would have to be compatible with `wasm-bindgen`, so we *would* have two `externref` tables. But it doesn't really apply, because we use the window handles to convert to types anyway, e.g. `JsValue` knows which table its implementation requires.

I propose we maintain the following properties:
- Handles must be an index to an `externref` table.
- Indices must be `usize`, to support Wasm64.
- Handles must be `!Send + !Sync`.

We should also change the following:
- Remove `NonNull`. This was specific to `wasm-bindgen`, but in general the null index is a valid entry into the `externref` table.
- Add an index for the table. A toolchain could potentially have multiple `externref` tables.

TLDR: what I said in #184 and introduce a second index for the table.

daxpedda:
So when I designed this proposal in my head, I was assuming that while the goal is toolchain neutrality, we still have to use the same toolchain on both ends of the handle.

The issue here is that in the future there most definitely will be multiple toolchains co-existing in the same binary, e.g. `wasm-bindgen` and the upcoming `js-bindgen`. They will have to co-exist because otherwise every dependency has to switch toolchain at the same time, making an ecosystem transition almost impossible. So yes, ultimately we do need a way to transfer `externref`s between toolchains and this is exactly what `raw-window-handle` should facilitate here.

[...]

PR #205, opened 2026-02-28
Rename web handles

madsmtm:
Discussed in https://github.com/rust-windowing/raw-window-handle/issues/185.


src/web.rs
madsmtm: I just realized, an option that might make sense, even semantically, could be to include the transfer function _in the display handle_? WDYT?

daxpedda (approved):

src/web.rs
daxpedda: The last update was that its almost impossible to represent a transfer function properly, regardless of where its stored.

Do you have a specific representation in mind?

Issue #188, opened 2026-02-21
Implement owned window handles using a Waker-type API


notgull:
In today's discussion, we discussed moving `raw-window-handle` to have an API similar to the [`Waker`] API in the standard library. Effectively, we'd have a cheaply cloneable "owned" window handle with facilities to increase and decrease its refcount. It would be implemented like so:

```rust
pub struct OwnedWindowHandle {
    data: NonNull<()>,
    vtable: &'static WindowHandleVTable
}

// OwnedWindowHandle is intentionally not Send and Sync.

pub struct WindowHandleVTable {
    // Decrease refcount.
    release: unsafe fn(NonNull<()>);
    // Increase refcount.
    retain: unsafe fn(NonNull<()>) -> Result<NonNull<()>, HandleError>;
    // Get the window handle.
    get_handle: unsafe fn(NonNull<()>) -> Result<WindowHandle<'_>, HandleError>;
}

impl OwnedWindowHandle {
    pub unsafe fn new(data: NonNull<()>, vtable: &'static WindowHandleVtable) -> Self { /* ... */ }
}

pub trait AsWindowHandle {
    fn as_handle(&self) -> Result<WindowHandle<'_>, HandleError>;
}

impl<T: AsWindowHandle + 'static + ?Sized> From<Rc<T>> for OwnedWindowHandle { /* ... */ }

pub struct ThreadSafeOwnedWindowHandle {
    inner: OwnedWindowHandle
}

unsafe impl Send for ThreadSafeOwnedWindowHandle {}
unsafe impl Sync for ThreadSafeOwnedWindowHandle {}

impl From<OwnedWindowHandle> for ThreadSafeOwnedWindowHandle { /* ... */ }
impl ThreadSafeOwnedWindowHandle {
    pub unsafe fn new(o: OwnedWindowHandle) -> Self { /* ... */ }
}
```

[`Waker`]: https://doc.rust-lang.org/std/task/struct.Waker.html

Effectively, rather than using `HasRawWindowHandle`, consumers would accept `Into<OwnedWindowHandle>` or `Into<ThreadSafeOwnedWindowHandle>`, depending on their requirements. Then, the consumer APIs could very easily just wrap around `OwnedWindowHandle` without imposing any lifetime or type requirements on their type. Meanwhile, on the producer, you could very easily implement `impl From<&winit::Window> for OwnedWindowHandle`  to use it with any graphics API.

[...]

PR #206, opened 2026-03-01
Make handles reference-counted

madsmtm:
Fixes https://github.com/rust-windowing/raw-window-handle/issues/188.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -131,7 +131,7 @@ pub enum WindowHandle<'window> {
     /// ## Availability Hints
     /// This variant should be expected anywhere Wayland works, which is
     /// currently some subset of unix systems.
-    Wayland(WaylandWindowHandle<'window>),
+    Wayland(WaylandWindowHandle),
     /// A raw window handle for the Linux Kernel Mode Set/Direct Rendering Manager
     ///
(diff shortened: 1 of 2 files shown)


src/appkit.rs
```diff
@@ -50,75 +51,214 @@ impl DisplayHandle<'static> {
...
+    pub unsafe fn new2(ns_view: NonNull<c_void>) -> Self {
+        #[link(name = "objc", kind = "dylib")]
+        unsafe extern "C-unwind" {
+            fn objc_retain(obj: *mut c_void) -> *mut c_void;
+            fn objc_release(obj: *mut c_void);
+        }
+
+        unsafe { Self::new(ns_view, objc_retain, objc_release) }
+    }
+
```
madsmtm: Unsure whether we should be in the business of exposing target-specific helper methods like these?

We definitely shouldn't make the `AppKitWindowHandle` struct be `#[cfg(target_os = "macos")]`, because it can technically also be used on Linux with GNUStep.

I also just now realized that this links `libobjc`, even if never used (which would be annoying for e.g. someone running Wayland on macOS). So if we wanted to do this, we might need to gate it behind a feature flag?

PR #207, opened 2026-03-01
Add reference counted handles

### LINKED ISSUE #188: Implement owned window handles using a Waker-type API

notgull:
Fixes #188

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,6 +47,7 @@ mod drm;
 mod gbm;
 mod haiku;
 mod ohos;
+mod owned;
 mod redox;
 mod uikit;
 mod wayland;
@@ -61,6 +62,10 @@ pub use drm::{DrmDisplayHandle, DrmWindowHandle};
 pub use gbm::{GbmDisplayHandle, GbmWindowHandle};
 pub use haiku::{HaikuDisplayHandle, HaikuWindowHandle};
 pub use ohos::{OhosDisplayHandle, OhosNdkWindowHandle};
+pub use owned::{
+    DisplayHandleVtable, OwnedDisplayHandle, OwnedWindowHandle, SyncDisplayHandle,
+    SyncWindowHandle, WindowHandleVtable,
+};
 pub use redox::{OrbitalDisplayHandle, OrbitalWindowHandle};
 pub use uikit::{UiKitDisplayHandle, UiKitWindowHandle};
 pub use wayland::{WaylandDisplayHandle, WaylandWindowHandle};
@@ -454,6 +459,13 @@ mod tests {
         assert_not_impl_any!(WasmBindgenOffscreenCanvasWindowHandle: Send, Sync);
         assert_impl_all!(AndroidNdkWindowHandle: Send, Sync);
         assert_impl_all!(HaikuWindowHandle: Send, Sync);
+
+        assert_impl_all!(OwnedDisplayHandle: Unpin, UnwindSafe, RefUnwindSafe);
+        assert_not_impl_any!(OwnedDisplayHandle: Send, Sync);
+        assert_impl_all!(SyncDisplayHandle: Unpin, UnwindSafe, RefUnwindSafe, Sync, Send);
+        assert_impl_all!(OwnedWindowHandle: Unpin, UnwindSafe, RefUnwindSafe);
+        assert_not_impl_any!(OwnedWindowHandle: Send, Sync);
+        assert_impl_all!(SyncWindowHandle: Unpin, UnwindSafe, RefUnwindSafe, Sync, Send);
     }
 
     #[allow(unused)]
--- a/src/owned.rs
+++ b/src/owned.rs
@@ -1,0 +1,565 @@
+//! Owned window and display handles.
+
+use crate::borrowed::{DisplayHandle, HasDisplayHandle, HasWindowHandle, WindowHandle};
+use crate::{HandleError, RawDisplayHandle, RawWindowHandle};
+
+use core::fmt;
+use core::ptr::NonNull;
+
+macro_rules! handle_type {
+    (
+        $name: ident,
+        $sync: ident,
+        $vtable: ident,
+        $raw: ident,
+        $borrowed: ident,
+        $as_borrowed: ident,
+        $as_borrowed_method: ident,
+        $type: literal
+    ) => {
+        #[doc = concat!("An owned ", $type, " handle.")]
+        ///
+        /// This type represents an owned pointer to a reference counted
+        #[doc = concat!($type, " handle. The underlying ", $type, " is expected")]
+        /// to stay alive as long as this type exists, and must not be closed
+        #[doc = concat!("or deallocated until all `", stringify!($name), "`s")]
+        /// produced from it have been dropped.
+        ///
+        /// In other words, this structure is effectively a type-erased
+        #[doc = concat!("[`Rc<dyn ", stringify!($as_borrowed), ">`](crate::", stringify!($as_borrowed), ").")]
+        /// On some platforms, it's possible to separate the underlying producer
+        #[doc = concat!("of a ", $type, " from the actual backing pointer.")]
+        ///
+        #[doc = concat!("The `", stringify!($name), "` consists of two main")]
+        /// parts. The first is a data pointer that represents the underlying
+        /// allocation or index of the handle object. The second is a virtual
+        /// function table pointer that describes how to interpret the
+        /// aforementioned pointer. This virtual function table currently contains
+        /// three functions:
+        ///
+        #[doc = concat!(" - `get`, which returns the underlying [`", stringify!($raw), "`]")]
+        ///   that represents the handle.
+        #[doc = concat!(" - `retain`, which increments the refcount of the ", $type, ".")]
+        #[doc = concat!(" - `release`, which decreases the refcount of the ", $type)]
+        ///   and deallocates it if the refcount drops to zero.
+        ///
+        /// It is expected that the handle is already `retain`ed before it is
+        /// created. In practical terms, if `retain` is called `N` times,
+        /// `release` will be called `N + 1` times. At the final invocation,
+        #[doc = concat!("the handle and ", $type, " will be deallocated.")]
+        ///
+        /// `retain`ing a handle and then immediately `release`ing the handle
+        /// should have no side effects. The `get` function is allowed to be
+        /// unpure and have side effects.
+        ///
+        /// This handle can be created in an unsafe way, from the pointer and
+        #[doc = concat!("vtable, using the [`from_raw_parts`](", stringify!($name), "::from_raw_parts)")]
+        /// method. Alternatively, this type can be created safely as well. It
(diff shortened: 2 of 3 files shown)


notgull:
This is meant to be an implementation of "outer reference counting", as opposed to #206, which is inner reference counting.

src/owned.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///   remain valid (i.e. not freed).
+    /// - The window handle returned by `get` must be valid for the lifetime of
+    ///   this window handle.
+    /// - `retain` and `release` must be atomic. This is necessary to ensure that
+    ///   `WindowHandle` is `Send` and `Sync`.
+    ///
+    /// Note that this API expects the underlying handle to be retained by
+    /// default. So if the `retain` callback is called `N` times, `release`
+    /// will be called `N+1` times. The underlying window must not be destroyed
+    /// as long as this reference is alive.
```
daxpedda: We have this part twice, once under `# Safety` and one more time here. I'm in favor of removing it from `# Safety`, as that should not be unsound (AFAIU).

Probably should be in its own paragraph here.

src/owned.rs
```diff
@@ -0,0 +1,549 @@
...
+    ///
+    /// # Safety
+    ///
+    /// - `ptr` must be a valid window object.
+    /// - `retain` must increment its refcount and `release` must decrement is
+    ///   refcount.
+    /// - While a single `retain`ed window handle still exists, the window must
+    ///   remain valid (i.e. not freed).
+    /// - The window handle returned by `get` must be valid for the lifetime of
+    ///   this window handle.
```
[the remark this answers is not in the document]
notgull: This is correct; the underlying v-table function is named `get`.

src/borrowed.rs
```diff
@@ -218,18 +218,18 @@ impl<H: HasWindowHandle + ?Sized> HasWindowHandle for alloc::sync::Arc<H> {
 /// See individual handle types for thread safety documentation. Since some
 /// window handle types are `!Send` and `!Sync`, this sum type is as well.
 #[derive(PartialEq, Eq, Hash, Copy, Clone)]
-pub struct WindowHandle<'a> {
+pub struct BorrowedWindowHandle<'a> {
```
madsmtm: Should be `WindowRef<'_>` IMO, and ideally done in a separate PR.

src/owned.rs
```diff
@@ -0,0 +1,547 @@
...
+}
+
+impl WindowHandle {
+    /// Create a new window handle from its raw parts.
+    ///
+    /// # Safety
+    ///
+    /// - `ptr` must be a valid window object.
+    /// - `retain` must increment its refcount and `release` must decrement is
+    ///   refcount.
```
madsmtm: Nit: spelling ("is refcount") and refer to what "its" is.

```suggestion
    /// - `retain` must increment the pointer's refcount and `release` must decrement it.
```

src/owned.rs
```diff
@@ -0,0 +1,547 @@
...
+use crate::borrowed::{
+    AsDisplayHandle, AsWindowHandle, BorrowedDisplayHandle, BorrowedWindowHandle,
+};
+use crate::{HandleError, RawDisplayHandle, RawWindowHandle};
+use core::ptr::NonNull;
+
+/// An owned window handle.
+pub struct WindowHandle {
+    /// Underlying data representative of the window handle.
+    ptr: NonNull<()>,
```
madsmtm: What's the reasoning for `NonNull<()>` here? `RawWaker` seems to use `*const ()`.

I guess if people aren't using it, they can always pass `NonNull::dangling()`, so it might be nice for niche optimizations?

src/owned.rs
```diff
@@ -0,0 +1,547 @@
...
+                NonNull::new_unchecked(alloc::sync::Arc::into_raw(value).cast::<()>() as *mut ()),
+                &Arc::<T>::VTABLE,
+            )
+        }
+    }
+}
+
+/// Virtual function table for [`WindowHandle`].
+pub struct WindowVtable {
+    /// Increment the refcount of the window handle.
```
madsmtm: Nit: These docs aren't shown to the user, put them in a comment on `new` instead.

src/owned.rs
```diff
@@ -0,0 +1,547 @@
...
+        Self {
+            retain,
+            release,
+            get,
+        }
+    }
+}
+
+/// An owned Display handle.
+pub struct DisplayHandle {
```
madsmtm: Q: Should we wrap things in a macro, to deduplicate the logical across window and display handles?

I'm leaning towards no, because I suspect we'll want them to behave a bit differently, but it'd be useful to discuss.

notgull: Leaning towards a macro.

src/owned.rs
notgull: No real point aside from have an extra point of nicheing. The handle type is already niched because of the `&'static VTable`. Frankly I could go either way; I just don't see the point of `null()` when `dangling()` exists.

notgull: My current opinion is that removing the possibility of the handle to be null potentially saves more footguns than allowing it.

src/lib.rs
```diff
@@ -449,8 +451,13 @@ mod tests {
         assert_not_impl_any!(WebOffscreenCanvasWindowHandle: Send, Sync);
         assert_impl_all!(AndroidNdkWindowHandle: Send, Sync);
         assert_impl_all!(HaikuWindowHandle: Send, Sync);
+
+        assert_impl_all!(WindowHandle: Send, Sync);
+        assert_impl_all!(DisplayHandle: Send, Sync);
+        assert_impl_all!(WindowVtable: Send, Sync);
+        assert_impl_all!(DisplayVtable: Send, Sync);
```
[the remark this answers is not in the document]
madsmtm: Rebasing seems to have accidentally removed these checks for the new types btw?

And I agree that we should check these in a test, regardless of whether we want the impls or not.

src/owned.rs
```diff
@@ -0,0 +1,613 @@
...
+            ///   used with the active virtual function table.
+            /// - `release` must decrement the handle's reference count.
+            /// - While a single `retain`ed handle exists, the
+            #[doc = concat!("   ", $type, " must remain valid (i.e. not freed or deallocated.")]
+            #[doc = concat!("  - In this rule, the term \"", $type, "\" refers to")]
+            ///      the underlying system resource.
+            #[doc = concat!(" - The [`", stringify!($raw), "`] must be valid for")]
+            ///   the lifetime of this handle.
+            /// - `retain`ing a handle and then immediately `release`ing the handle
+            ///   must have no side effects. The `get` function is allowed to be
```
madsmtm:
```suggestion
            ///   should have no side effects. The `get` function is allowed to be
```

And maybe move elsewhere, it's not actually a needed safety requirement IIUC?

src/owned.rs
```diff
@@ -0,0 +1,613 @@
...
+            /// - The data pointer must be a [valid](https://doc.rust-lang.org/std/ptr/index.html#safety)
+            #[doc = concat!("   pointer to an object representing a ", $type, ".")]
+            /// - `retain` must increment the handle's reference count. The pointer
+            ///   returned by `retain` should be another valid pointer that can be
+            ///   used with the active virtual function table.
+            /// - `release` must decrement the handle's reference count.
+            /// - While a single `retain`ed handle exists, the
+            #[doc = concat!("   ", $type, " must remain valid (i.e. not freed or deallocated.")]
+            #[doc = concat!("  - In this rule, the term \"", $type, "\" refers to")]
+            ///      the underlying system resource.
```
madsmtm: A bit unclear, does "in this rule" refer to the text above? If so, maybe just merge the two bullet points?

src/owned.rs
```diff
@@ -0,0 +1,613 @@
...
+            #[doc = concat!(" - The [`", stringify!($raw), "`] must be valid for")]
+            ///   the lifetime of this handle.
+            /// - `retain`ing a handle and then immediately `release`ing the handle
+            ///   must have no side effects. The `get` function is allowed to be
+            ///   unpure and have side effects.
+            ///
+            #[doc = concat!("See the [structure level description](", stringify!($name), ")")]
+            /// for further elaboration on the expectations for each vtable method.
+            // TODO: Add examples
```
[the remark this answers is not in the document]
notgull: This could be added later on.

src/owned.rs
```diff
@@ -0,0 +1,591 @@
...
+                    }
+
+                    #[inline(always)]
+                    unsafe fn get(ptr: NonNull<()>) -> Result<$raw, HandleError> {
+                        let value: &T = unsafe { ptr.cast().as_ref() };
+                        value.$as_borrowed_method().map(Into::into)
+                    }
+                }
+
+                // SAFETY: Above vtable is valid for this Rc
```
daxpedda:
```suggestion
                // SAFETY: Above vtable is valid for this `Rc`.
```

PR #209, opened 2026-03-01
Make `Win32WindowHandle` `Send + Sync`

### LINKED ISSUE #196: Making Windows `Send + Sync`
Currently we have the handle on windows `!Send + !Sync` because some APIs are not thread safe to use.
However, the whole rendering API is thread-safe.

Instead of blocking this, we should encode the `ThreadId` in the handle and consumers have to figure out themselves if what they are doing is safe or not.

daxpedda:
Resolves #196.

### DIFF
--- a/src/windows.rs
+++ b/src/windows.rs
@@ -49,10 +49,10 @@ impl DisplayHandle<'static> {
 ///
 /// ## Safety
 ///
-/// Note that window objects have [thread affinity]. Not all functions of the
-/// Win32 handle are thread-safe (modifying functions especially), so care
-/// should be taken to not call these functions from other threads. When in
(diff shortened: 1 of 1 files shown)


madsmtm:
Also, would it make sense to wait for https://github.com/rust-windowing/raw-window-handle/issues/188? Since the safety of this depends on what `Win32WindowHandle` ends up being. (If we choose option A in that thread, the handle would become `!Send` + `!Sync` again).

Then again, I'm also fine with doing this now (it's correct for the handle as currently defined), and then worry it this later when we do get around to changing the semantics of handles.

See also https://github.com/rust-windowing/raw-window-handle/pull/192, that one has a bit more detailed docs.

Issue #178, opened 2026-02-06
Merge `RawWindowHandle` and `WindowHandle<'_>`?


madsmtm:
I don't recall the reasoning for keeping `RawWindowHandle` around, is there ever a case where it's useful? And wouldn't `WindowHandle<'static>` be able to serve that same purpose?

I propose we change our API to merge these, so that it's instead something like:
```rust
#[non_exhaustive]
enum WindowHandle<'window> {
    AppKit(AppKitWindowHandle<'window>),
    UIKit(UIKitWindowHandle<'window>),
    AndroidNDK(AndroidNDKWindowHandle<'window>),
    // ... etc.
}
```

This means that instead of a three-step process for making safe handles (create raw platform handle + wrap in `RawWindowHandle` + call `WindowHandle::borrow_raw`), it's now only two steps (create raw platform handle with lifetime + wrap it in `WindowHandle`).

One reason to not do this would be to mirror the `std::os::fd` API (`RawFd` vs `BorrowedFd<'_>`), but we already can't do that, because we can't provide a mirror of `OwnedFd`.

notgull:
I'm fine with getting rid of raw window handles entirely, assuming we can't come up with a better use case. As an added bonus, it lets us have more granular safety rules for each window handle. For example, we can make it safe to construct an XCB window handle, but not a Win32 window handle.

One argument for raw handles is it lets us have a raw pointer equivalent for window handles.

notgull:
Two points from today's meeting:

- It's good to have a `RawWindowHandle` that represents plain old data, compared to a `BorrowedWindowHandle<'static>` that represents a valid live window handle.
- We should move forwards with getting rid of the subtypes.

Issue #195, opened 2026-02-28
Remove the word "handle" from the end of all of the types


notgull:
Basically every type ends with the word "Handle".

<img width="268" height="672" alt="Image" src="https://github.com/user-attachments/assets/813cbdcb-e932-4b50-902a-166cebe5b607" />

We could probably save a lot of typing time and make the names a lot nice by removing the "handle" at the end. E.g. `WindowHandle` would become `Window`, `AndroidNdkWindowHandle` would just become `AndroidNdkWindow`, etc.

madsmtm:
I guess what I was hoping for from https://github.com/rust-windowing/raw-window-handle/pull/187 to start with would be that we rename `Window` to `Surface`. I'd be fine going all the way with renaming `WindowHandle` -> `Surface` too.

Then again, `Handle` maybe makes sense if our surfaces are analogous to weak references (see https://github.com/rust-windowing/raw-window-handle/issues/188#issuecomment-3976393331)?

Lokathor:
I like to fully import types, and so I naturally end up thinking that single-word types only really work well when the word won't be "claimed" by multiple crates commonly used together. However, since this crate is *inherently* about being glue code, we can assume that winit and sdl and whatever else would reasonably have their own types named `Window` and perhaps `Display` and perhaps `Surface` and perhaps `Context`, which would all interfere with our type names, forcing people to prefix something.

So even though it makes sense in the context of *just* our crate to remove the "handle" part, I'm not sure it's as good idea for the end users.

PR #214, opened 2026-03-14
Merge platform-specific handles into `Raw*Handle`

madsmtm:
Implements the idea discussed in https://github.com/rust-windowing/raw-window-handle/issues/178#issuecomment-3858254821.

Benefits:
- Fewer types, the API is simpler for users and the docs are less cluttered.

Cons:
- The thread-safety guarantees that was figured out in https://github.com/rust-windowing/raw-window-handle/pull/192 and https://github.com/rust-windowing/raw-window-handle/pull/209 are now documentation-only (since there isn't a type to attach the `Send`/`Sync` impl to). Dubious of how much value this was to begin with though.
- Bit harder to set fields like `hinstance` or `visual_id` afterwards.
  - Could be alleviated with extra helper methods (or making the type exhaustive).
- Platform handles are split between files instead of being defined "together".
- Fields have to be public (unless we re-add wrappers in specific cases where needed).

Fixes https://github.com/rust-windowing/raw-window-handle/issues/178.
Replaces https://github.com/rust-windowing/raw-window-handle/pull/204 and https://github.com/rust-windowing/raw-window-handle/pull/194.
Makes https://github.com/rust-windowing/raw-window-handle/issues/195 less of a problem.


src/lib.rs
```diff
@@ -40,270 +40,16 @@ extern crate alloc;
...
-mod haiku;
-mod ohos;
-mod redox;
-mod uikit;
-mod wayland;
-mod web;
-mod windows;
-mod x11;
+mod raw_display;
+mod raw_window;
```
madsmtm: The diff is a bit smaller if we don't do the file move, but I felt it made sense to do now if we want that?

Since the diff in the squashed commit will be simpler.

src/raw_window.rs
```diff
@@ -0,0 +1,863 @@
...
+    ///
+    /// let value: OffscreenCanvas;
+    /// # value = todo!();
+    ///
+    /// // Convert to the raw index and convert to the handle.
+    /// let index = (&value).into_abi();
+    /// let handle = RawWindowHandle::new_weboffscreencanvas(index as usize);
+    /// # }
+    /// ```
+    pub fn new_weboffscreencanvas(obj: usize) -> Self {
```
madsmtm: Naming scheme? I don't like this name, but `new_app_kit` / `new_ui_kit` also sounds kinda weird (though I guess less weird).

src/raw_display.rs
```diff
@@ -0,0 +1,515 @@
...
+impl RawDisplayHandle {
+    /// Create a new empty display handle.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use raw_window_handle::RawDisplayHandle;
+    /// let handle = RawDisplayHandle::new_uikit();
+    /// ```
+    pub fn new_uikit() -> Self {
```
madsmtm: Do we need this when we have `WindowHandle::uikit()`?

notgull: I would say yes.

madsmtm: See also https://github.com/rust-windowing/raw-window-handle/issues/215.

PR #217, opened 2026-03-16
Rename `DisplayHandle` convenience methods

madsmtm:
In https://github.com/rust-windowing/raw-window-handle/pull/70#issuecomment-855261670, we decided upon a naming scheme for our types, with the main motivation being "consistency" and "`AndroidNDKHandle` looks weird"[^1].

[^1]: This argument is weakened after https://github.com/rust-windowing/raw-window-handle/pull/214 where this type no longer exists. Now it's only `RawWindowHandle::AndroidNdk` vs. `RawWindowHandle::AndroidNDK`.

I'm still mostly convinced that this is still the right direction to go in because it matches things like `std::net::Ipv4Addr` (if it was `IPv4Addr`, then we should've done something different).

We should apply this naming scheme to methods as well, in particular `DisplayHandle::appkit` and `DisplayHandle::uikit` need an extra `_` to follow normal case-conversion rules.


notgull:
I think `ui_kit` works weird, so I'd advocate for `uikit`.

kchibisov:
The crate names we use in winit are `appkit`/`uikit`, so I'd probably have the names without `_` here, so it kind of the same spelling.

PR #221, opened 2026-03-24
Implement `core::error::Error` when `std` feature is disabled

madsmtm:
This requires a bit higher MSRV, which we document.

See discussion in https://github.com/rust-windowing/keyboard-types/pull/75 and https://github.com/rust-windowing/keyboard-types/pull/70 for reasoning (basically: MSRV shouldn't hold the nice API back for `no_std` people, and the overlap between `no_std` + low MSRV + windowing is really low).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,7 +37,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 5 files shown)


daxpedda (approved):

kchibisov (approved):
I've bumped MSRV to 1.85 to raise edition to 2024, since we do breaking release anyway no point in holding it.

